PR #64, opened 2014-07-03
Add and Remove point by ID

chainlink:
Can now add or remove a point by id
Related to #49


chrisn:
Adding/removing points by ID is now resolved by #141. But this PR contains an additional fix for point colours, which I'll raise as a separate issue.

Issue #87, opened 2014-09-25
Markers API


oncletom:
# tl;dr

This API replaces the Segments and Points API.
# Why?

Segments and points contains basically the same data. They differ only in their rendering aspect.
Hence a very similar codebase.

Also, we want to deal with multiple/optional layers of data, to toggle them.
# Goals
- light object structure
- data to rendering approach
- a marker = timing informations + free structure (meta)data + rendering function
- view system to narrow the selection per waveform view

# Example

``` js
var p = Peaks.init({
  // ...
});

// add a marker
p.markers.add({
  timestamp: 30.5,
  duration: 20,
  data: { name: 'John Reith', url: 'http://dbpedia.org/page/John_Reith,_1st_Baron_Reith', labels: ['speaker'] }
});

// add another one which is non-editable, with a custom shape
p.markers.add({
  timestamp: 30,
  editable: false,
  data: { comment: '...', pid: 'b01mxx3h' }
  shape: (segment, view) => {
    const shape = new Konva.Shape({ ... });
    shape.sceneFunc(...);

   if (view.name === 'overview') {
     return false;
   }

    return shape;
  },
});

// add views ...
p.markers.views.speakers = (marker) => marker.data.labels.indexOf('speaker') !== -1);
p.markers.views.comments = (marker) => 'comment' in marker.data;

// fetched by view
p.markers.views.speakers.forEach(marker => { ... });

// all of them
p.markers.forEach(marker => { ... });

// change some data in markers and force update their rendering
p.on('marker.click', (event, view, markers) => {
  markers
    .filter(view.objectInView)
    .forEach(marker => { ... });

  markers.render();
});
```
# Work
- [ ] Definition of the features
- [ ] Definition of the API

closes #64

oncletom:
@chrisn I amended the API definition to make it more lightweight and able to fit in people existing data structures, possibly by referencing data rather than having to duplicate them in memory.

What do you think?

chrisn:
Looks good, a few thoughts below:
* Another goal is to allow the number of views to be configurable (https://github.com/bbc/peaks.js/issues/2).
* Specifing `startTime` and `endTime` for segments is preferable to `time` and `duration`, although for convenience we could allow both. I also suggest renaming `timestamp` to `time` for point markers, for consistency.
[...]

PR #89, opened 2014-09-27
Allow configuration of where segments are displayed

lowpitch:
This commit adds initialization options to peaks, allowing you to specify whether segments should appear in the Zoom View, the Overview, or both.

Reference issue #89 

The default behaviour is for segments to appear in both views, so no existing code should be affected by this addition.

The following initialization object will tell peaks to only show segments in the Zoom View, and not in the Overview.

```
var options = {
  container: document.getElementById('first-waveform-visualiser-container'),
  mediaElement: document.querySelector('audio'),
  dataUri: {
    arraybuffer: '/test_data/sample.dat'
  },
  showSegmentsInZoomView: true,
  showSegmentsInOverview: false
};

var peaksInstance = Peaks.init(options);
```

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -118,7 +118,13 @@ define('peaks', [
        */
       pointMarkerColor:     '#FF0000', //Color for the point marker
       pointDblClickHandler: null, //Handler called when point handle double clicked.
-      pointDragEndHandler:  null // Called when the point handle has finished dragging
+      pointDragEndHandler:  null, // Called when the point handle has finished dragging
+
+      /**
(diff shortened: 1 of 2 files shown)


oncletom:
@chrisn this should be closed and superseded by #87

chrisn:
Sorry it's taken so long to get round to this... https://github.com/bbc/peaks.js/commit/53d3e6491418f7c9943729cabc220ded5aa2fd47 adds `enableSegments` and `enablePoints` options to completely disable rendering of points and segments in each view. Separately, we could add an API to toggle them on or off,

Issue #69, opened 2014-07-22
Method to destroy peaksInstance


bitwit:
So we are currently using peaks.js in the context of a single page application and have situations where multiple peaksInstances can be created before a page reloads.

What we are seeing is that after loading the second peaks instance, any browser resizing crashes the tab. Note, we can load many peaks instances on top without issue but the browser resizing is what causes the crash. 

We suspect that we are creating memory leaks by not properly destroying a peaksInstance before replacing it and there are perhaps lingering callbacks that should be deregistered.

So this is sort of a feature request/request for guidance on how we might go about making sure peaks has completely cleared. There appears to be no peaks.destroy() method.

Much appreciated!

oncletom:
@chrisn I guess this one can be closed since #150 was merged.

PR #92, opened 2014-09-29
Adding 'refresh' function to peaks instance

chainlink:
Switched order of src and currentSrc.
(src is not set when setting audio via source tags in html. CurrentSrc
is not current when resetting src through javascript.)

See #69

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -133,6 +133,11 @@ define('peaks', [
      * @type {number}
      */
     this.currentZoomLevel = 0;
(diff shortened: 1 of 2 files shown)


chrisn:
https://github.com/bbc/peaks.js/commit/3cf5fee268edf5dcae4cf6d3ce777b8d64b7cf5e adds a `setSource()` function that allows you to change the media URL and refresh the waveform.

PR #95, opened 2014-09-30
Make point editable in zoom view only.

chainlink:
screen shot 2014-09-28 at 12 04 50 pm

When the overview is sufficiently sized, the handle appears, but is not veritically centered to height. We think that it's also clutter to have the the drag handle in both places. This change makes points only editable in the zoom view.

Note: The demo commit can be omitted, as it only changes the height of the overview in the index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,7 +21,7 @@
       }
 
       .overview-container {
-        height: 85px;
+        height: 100px;
       }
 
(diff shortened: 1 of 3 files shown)


chrisn:
Closing, as point and segment markers are editable by default in the zoomable view and not editable in the overview waveform. The `view.enableMarkerEditing()` method allows this to be changed, if desired.

Issue #105, opened 2014-11-12
sticky backward seeking and undraggable segment handles


un-chein-andalou:
Hi folks,

I just started playing with peaks and find it's mostly working brilliantly! but a couple of things..

 I don't know why but backward seeking is kinda sticky:
When clicking far ahead of the playhead in the overview view the zoom view jumps right to the correct spot -awesome, but when clicking far behind the play head in the overview view the zoom view doesn't move all the way to the correct spot -less-awesome. With a big sized audio file you have to do a lot of clicking on the overview to move the zoom view all the way back so this can be annoying.

also I wonder why dragging segments is disabled in the overview:
in waveform.segments.js ln 68:70
```javascript
          if (segmentGroup === segmentOverviewGroup) {
            draggable = false;
          }
```
I guess dragging segments in the overview was disabled for a reason, but still I need to drag segment handles around in the overview so I went ahead and enabled it. Will it cause problems? I haven't seen any issues so far but thought I'd check.

oncletom:
Hello @un-chein-andalou, sorry for the delay and thank you for sharing your findings. Any chance you can propose that as a PR? Thanks :-)

un-chein-andalou:
created PR #108

oncletom:
Thanks @un-chein-andalou, I will try to give it a go next week :-)

oncletom:
Solved by #108

PR #108, opened 2015-01-11
fix for bug #105

un-chein-andalou:
:)


oncletom:
Thanks @un-chein-andalou! I will remove the segment id fix though, as it should be addressed differently (either through a uniquid or time based function).

PR #109, opened 2015-01-14
Add event for handle being dragged

samstarling:
I would find it useful for there to be an event that's fired when a segment handle is dragged. I couldn't see events being tested anywhere, so i haven't added a test. I hope I'm contributing this in a sensible way, and in the correct place! Let me know if not.

### DIFF
--- a/src/main/markers/waveform.segments.js
+++ b/src/main/markers/waveform.segments.js
@@ -155,6 +155,8 @@ define([
         var outOffset = thisSeg.view.frameOffset + thisSeg.outMarker.getX();
         segment.endTime = thisSeg.view.data.time(outOffset);
       }
+      
+      peaks.emit("segments.handle_dragged");
 
       updateSegmentWaveform(segment);
     };


oncletom:
Hello @samstarling :-)

Thanks for the contribution! What would you had the ability to listen to a `segments.handle_dragged` dragged?

IMO, we should at least provide the `segment` object as an argument, so as people could now which segment it is about, with its updated start/end properties.

Agreed it is difficult to test — ideally, it implies to see if a call to Kinetic Segment API would trigger a call to this Peaks instance event, with the expected arguments.

un-chein-andalou:
hi @samstarling  @oncletom 
agree with oncletom, but for me I didn't need the whole segment object so I use:
```javascript
peaks.emit("segment.drag", [segment.id, segment.startTime, segment.endTime]);
```
meh,  works for me :)

samstarling:
Hey @un-chein-andalou - that sounds like a good solution, I'll try it out!

oncletom:
Thanks. Final implementation is `segment.dragged` with the `Segment` object. Cf. 17aa95fd9979394ace81833aad795ff146ffb8f9

PR #113, opened 2015-04-02
Make the axis label and gridline colors configurable

ziggythehamster:
This also fixes a regression from the KineticJS upgrade, which broke painting the axis in anything other than black. Using `setAttr` on the context makes it work as expected and now the axis is painted in the correct color.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   "license": "LGPL 3",
   "scripts": {
     "build": "browserify -d ./src/main.js -p [minifyify --map peaks.min.map --output peaks.min.map] -s peaks -o peaks.min.js",
+    "build-max": "browserify ./src/main.js -s peaks -o peaks.js",
     "test": "npm run test-pre && ./node_modules/karma/bin/karma start --single-run",
     "test-pre": "jshint ./src",
     "test-watch": "./node_modules/karma/bin/karma start --auto-watch --no-single-run",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
   "license": "LGPL 3",
   "scripts": {
     "build": "browserify -d ./src/main.js -p [minifyify --map peaks.min.map --output peaks.min.map] -s peaks -o peaks.min.js",
+    "build-max": "browserify ./src/main.js -s peaks -o peaks.js",
```
oncletom: What is `build-max` done for?

oncletom:
Thanks a lot. If I understand well, `context` would not pick the change unless `setAttr` is called?

package.json
ziggythehamster: I take the outputted `peaks.js` file and drop it into a Broccoli pipeline which minifies it and inlines it with my other JavaScript. It does not work very well when it tries to minify the already minified version.

ziggythehamster: Also, I didn't intend for that commit to make it to this PR, but it doesn't hurt anything for it to be there ;)

ziggythehamster:
Correct - `context` in this case is actually a Kinetic/Konva object rather than a native canvas, and setting the property doesn't actually do anything unless you use Kinetic/Konva's API to set it (`setAttr`).

oncletom:
:+1:

package.json
oncletom: It does not, but it does not help to carry over more and more things without a need.

I agree though building a non-minified artefact can be useful :-)

Issue #114, opened 2015-04-03
Zoom causes a browser freeze if done after waveformOverviewReady


ziggythehamster:
Tested browsers:
* Safari 7.1.3 (OS X Mavericks)
* Safari 7.1.4 (OS X Mavericks)
* Firefox 38 (OS X Mavericks)
* Opera Developer 30.0.1820.0 (OS X Mavericks)
* IE 11 (Windows 8.1)
* IE 11 (Windows 10)

I haven't tried it with Project Spartan, but I have no reason to believe that it will work there.

If I have the following init code (where peaks is `window.peaks` because Browserify doesn't work as expected with ES6 modules and I don't want to figure out why):

```coffeescript
    $mediaElement = $("audio")
    p = peaks.init
      container: $(".waveform-container").get(0)
      height: 200
      logger: console.error.bind(console)
      mediaElement: $mediaElement.get(0)
      segments: []
      zoomLevels: [256, 512, 2048, 8192]

      dataUri:
        arraybuffer: "https://SOME_BUCKET.s3.amazonaws.com/SOME_FILE.dat"

    p.zoom.setZoom(2)
```

Then zoom works as expected: I am now zoomed at the 2048 zoom level.

If I zoom after `waveformOverviewReady` fires, the browser locks up. If I don't zoom like I'm doing here, it still locks up. The demo that is currently in master under `npm run start` freezes as well. I was able to get a LITTLE bit of profiling information from IE (and only IE) and it seems like it's freezing in `Kinetic.Animation._animate`. All I see is a bunch of garbage collection for tens of seconds (and then I stopped the profiler). Maybe it's stuck in an infinite loop?

I think this is also why the build is failing to run properly under CI - perhaps it times out because the browser freezes.

What should I try to diagnose this? Is there something I can configure or comment out to make it zoom without animating?

lowpitch:
ello,

I've also experienced this locking-up, and the cause seems to be a bug in the version of the Kinetic node module that's pulled in.

If you're building your own version of peaks.js, a temporary workaround is to make a change in the kinetic.js file to fix the bug, then to rebuild. For me, around line #10184 there's a function called getIntersection - this is where the infinite loop occurs. There's a block of code which looks like:

```javascript
// if no shape, and no antialiased pixel, we should end searching
 if (continueSearch) {
  spiralSearchDistance += 1;
} else {
  return;
}
```

Change this to

```javascript
// if no shape, and no antialiased pixel, we should end searching
[...]

PR #118, opened 2015-06-24
Convert the library to use Konva.js instead of KineticJS and fix a bug with zooming

ziggythehamster:
This fixes the bug described by #114 as well as switches the library to use Konva.js without any deprecation warnings.

### DIFF
--- a/src/main/views/waveform.overview.js
+++ b/src/main/views/waveform.overview.js
@@ -183,6 +183,9 @@ define([
   };*/
 
   WaveformOverview.prototype.updateRefWaveform = function (time_in, time_out) {
+    if (isNaN(time_in))  throw new Error("WaveformOverview#updateRefWaveform passed a time in that is not a number: " + time_in);
+    if (isNaN(time_out)) throw new Error("WaveformOverview#updateRefWaveform passed a time out that is not a number: " + time_out);
+
     var that = this;
 
     var offset_in = that.data.at_time(time_in);
@@ -199,6 +202,8 @@ define([
   };
 
   WaveformOverview.prototype.updateUi = function (pixel) {
+    if (isNaN(pixel)) throw new Error("WaveformOverview#updateUi passed a value that is not a number: " + pixel);
+
     var that = this;
 
     that.playheadLine.setAttr("x", pixel);
--- a/src/main/views/waveform.overview.js
+++ b/src/main/views/waveform.overview.js
@@ -9,8 +9,8 @@
 define([
   "peaks/waveform/waveform.axis",
   "peaks/waveform/waveform.mixins",
-  "Kinetic"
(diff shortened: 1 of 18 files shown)


src/main/views/waveform.overview.js
```diff
@@ -183,6 +183,9 @@ define([
   };*/
 
   WaveformOverview.prototype.updateRefWaveform = function (time_in, time_out) {
+    if (isNaN(time_in))  throw new Error("WaveformOverview#updateRefWaveform passed a time in that is not a number: " + time_in);
```
oncletom: Is there any reason `time_in` and `time_out` to equal `NaN`? I don't see any code change in our code base which could cause this.

The question could be rephrased as *how useful is it*, because if I understand well, the bug lied in Kinetic.

oncletom:
Thanks @ziggythehamster, that's definitely an incredibly helpful piece of work.
The build fails so I will have a look at it prior to merging.

Has Konva a backward compatible API? I see some differences here and there.

Can I also ask you to rebase your work and to squash it as a single commit? (or 2 commits, one for the konva migration, one for the bugfix – one commit will be largely okay though).

@lowpitch how does it affect your current stream of work?

src/main/views/waveform.overview.js
ziggythehamster: There _shouldn't_ be any reason why these will ever be NaN, but the bug was extremely subtle and I only found it by checking the arguments are the correct type. Since you're expecting time_in and time_out to be numbers, I am asserting that they are at the top of the method. I should be doing these assertions everywhere in waveform.*.js, so if I missed one, let me know and I'll add it :)

ziggythehamster:
Konva's API is fully backwards compatible, much to my surprise (I budgeted this conversion to take the rest of the week!). There was only one thing that was deprecated, and I fixed those deprecation warnings by switching from `setDrawFunc` to `sceneFunc` which appear to do the exact same thing. It still worked with `setDrawFunc`, but it logged a warning telling you that it's deprecated. :)

I'll squash this into 3 commits:

* Konva.js changes
* Type checking arguments in waveform.*.js
* The actual bug fix

oncletom:
Great! Thank you very much :-)

ziggythehamster:
This has been squashed, ended up being 4 commits because I forgot your bower.json was wrong (since you switched to Browserify). I can make that a separate PR if you want.

oncletom:
@ziggythehamster some people still use bower (and install all the deps through bower) but I don't plan to support it in the future. npm is the way to go (and Kinetic was not available through npm for a long time). So don't bother — I'll probably major bump the version of the package as well as the bower support is removed.

ziggythehamster:
I updated this to go against latest master.

ziggythehamster:
And it is now updated once again to go against latest master :)

PR #121, opened 2015-06-25
Make more colors configurable

ziggythehamster:
This PR adds more configuration options for setting colors. Specifically:

* The overview highlight rectangle (if you have a better name, let me know). The grey blob in the overview.
* The text color of the playhead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,11 +182,17 @@ var options = {
   // Colour for the overview waveform
   overviewWaveformColor: 'rgba(0,0,0,0.2)',
 
+  // Colour for the overview waveform rectangle that shows what the zoom view shows
+  overviewHighlightRectangleColor: 'grey',
+
   // Colour for segments on the waveform
   segmentColor: 'rgba(255, 161, 39, 1)',
 
   // Colour of the play head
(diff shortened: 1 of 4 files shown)


ziggythehamster:
Don't merge this just yet, seems like I missed a couple.

ziggythehamster:
A point's passed color isn't honored anywhere, but it looks to be a pretty big undertaking to make that happen, so I'm not touching it right now :).

ziggythehamster:
@oncletom Is there anything else this needs in order to get merged?

oncletom:
@ziggythehamster well, nothing much really, I kept on the idea of the `Don't merge this just yet` ;-)

oncletom:
Actually reviewing the change again, you added `pointMarkerColor` in the README but it does not exist in the codebase so we should get rid of it unless if you plan to make the big change now.

ziggythehamster:
You already had pointMarkerColor, it just wasn't documented. The color property of points is specifically what doesn't look easy to fix - to get it to do the same thing as the segment color property.

oncletom:
:+1: oh okay, then we're good :-)

Issue #102, opened 2014-10-15
Jumpy playhead


martybalandis:
#99 worked from CDN and surprisingly I managed to upgrade from 0.0.X to 0.4.3 within 30min. I have a tiny annoying issue though.

Unlike in 0.0.X version, playhead in the latest version is feels really jumpy. Did I forget something imortant? In the old version it is really smooth, I don't seem to find anything in my code logically that could cause this.

This is the video that shows it: https://dl.dropboxusercontent.com/u/680402/peaks.mp4

This is what my code basically looks like:
            (function(Peaks){
                var options = {
                    container: document.getElementById('waveform-visualiser-container'),
                    mediaElement: document.getElementById('audio'),
                    dataUri: {
                    arraybuffer: datFile
                    },
                    keyboard: false,
                    height: uriParam("height"),
                    zoomWaveformColor: zoomWaveformColor,
                    overviewWaveformColor: overviewWaveformColor,
                    playheadColor: 'rgba(255, 250, 255, 1)'
                };
                Peaks.init(options);
            })(peaks.js);

I am using peaks.js file which I stored on my own server.

un-chein-andalou:
jumpy playhead!

 It starts out jumpy, but if one hits pause then play again - it runs smooth! But goes jumpy again after every seek.

chrisn:
This is fixed in [v0.9.0](https://github.com/bbc/peaks.js/releases/tag/v0.9.0). Feel free to re-open this issue if there's still a problem.

PR #122, opened 2015-06-26
Fix typo, seeking instead of seaking

ziggythehamster:
You have a typo in waveform.overview.js, this fixes it :)

### DIFF
--- a/src/main/views/waveform.overview.js
+++ b/src/main/views/waveform.overview.js
@@ -75,7 +75,7 @@ define([
     // EVENTS ====================================================
 
     function trackPlayheadPosition(time, frame){
-      if (!peaks.seaking) {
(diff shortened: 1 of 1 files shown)


oncletom:
You mean #102? Or another reported issue?

Thanks in any case :-)

ziggythehamster:
I have no idea what this affects, or even if it's been reported, but you obviously meant `seeking` :). This fix makes it so that the state is more consistent - otherwise, if you go crazy, you end up with maybe the playhead thinking it's at one position and then going past the end or beginning of the file, but it's actually at a different position.

oncletom:
Agreed, it is definitely `seeking`. I would have linked and closed the related issue otherwise!

Thanks a lot, definitely helpful :-)

PR #123, opened 2015-06-29
Allow alternate zoom adapters, add a static (non-animated) zoom adapter, add more safety checks

ziggythehamster:
We have some really large (3+ hours at 256 samples/pixel!) audio files where zoom is never really going to be performant, so we wanted to be able to turn zoom off.

This PR also adds safety checks (some of which are already in #118, see commit 1395290) to prevent doing things which are incorrect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,8 +197,8 @@ var options = {
   // Random colour per segment (overrides segmentColor)
   randomizeSegmentColor: true,
 
-  // Zoom view adapter to use. Valid adapters are: 'AnimatedZoomAdapter' (default) and 'StaticZoomAdapter'
-  zoomAdapter: 'AnimatedZoomAdapter',
+  // Zoom view adapter to use. Valid adapters are: 'animated' (default) and 'static'
+  zoomAdapter: 'animated',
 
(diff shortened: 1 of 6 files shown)


oncletom:
Hello,

thanks for your contributions, this is great to see people contributing back what peaks brings to them!
Slight suggestion which does not invalidate what you have done but simply make it modular: instead of having a boolean flag for animations, I would rather use a string one, coining which zoom adapter to use.

Have a look at `src/main/views/zooms` and create a non-animated one (with roughly the content of your `else` statement). Follow the same interface and use it instead of creating an `if/else` statement.

Let me know how you feel about it :-)

ziggythehamster:
@oncletom My colleague has updated his branch with your suggestions - what do you think? :)

oncletom:
Great job @johvet and @ziggythehamster :-) I have made a small comment to get rid of the `switch` statement.

ziggythehamster:
See latest changes :)

oncletom:
That's perfect :-) thanks a lot for these contributions!

Issue #53, opened 2014-06-27
Views API


oncletom:
Peaks should provide an easier way to add, remove and order views.

Using native views with default settings:

``` js
peaks.init({
  views: ['zoom', 'overview']
});
```

Using only one configured native views:

``` js
require(['peaks', 'peaks/views/zoom'], function(Peaks, ZoomView){
  var zoom = new ZoomView({
    height: 240,
    backgroundColor: '#fff',
    textColor: 'red'
  });

  var p = Peaks.init({
    views: [zoom]
  });
});
```

Combining a default native view and your custom metadata view:

``` js
require(['peaks', 'my/html/view'], function(Peaks, CustomView){
  var tableView = new CustomView({
    speakersDataUrl: '//localhost:3000/speakers/worldcup/france-vs-switzerland.json'
  });

  var p = Peaks.init({
    views: [tableView, 'overview']
  });
});
```
# View

A view is an asynchronous renderer, performing actions based on peaks timed interactions.

``` js
var ZoomView = Object.create(BaseView.prototype, {
  id: 'zoom',
  init: function(config, peaks){
    peaks.on('segment.add', this.onAddSegment.bind(this);

    peaks.emit('load.views.' + id, this);
  },

  onAddSegment: function(segment){

  },

  timeUpdate: function(time){

  },

  render: function(){

  }
});
```

closes #2 
closes #46
closes #59

chrisn:
This is resolved in release 0.10.0, please see the [demo](https://github.com/bbc/peaks.js/blob/master/demo) folder for examples.

PR #124, opened 2015-07-02
Add more configuration options to the overview waveform, make it possible to turn on or off the axes

ziggythehamster:
You can now configure:
- The highlight rectangle padding (we use this because we make our overview bar smaller)
- The highlight rectangle opacity
- The overview axis colors

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,9 @@ var options = {
   // Colour of the axis labels
   axisLabelColor: '#aaa',
 
+  // Use an axis on the following waveforms: 'overview', 'zoomview'
+  axisOnWaveforms: ['overview', 'zoomview'],
+
   // Random colour per segment (overrides segmentColor)
   randomizeSegmentColor: true,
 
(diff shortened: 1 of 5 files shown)


ziggythehamster:
This now goes against latest master :)

ziggythehamster:

oncletom:
Hey!

Thanks for the proposals, they are valuable options!

The list of options is now very long, and I wonder how effective it is for newcomers and peaks users.

In issue #53, I propose to use the same config objects but to pass them to the views directly in order to control them. Next step would be to use the same objects for both views (zoom and overview) and to affect their layout only via configuration. Peaks responsibility would then be to synchronise/handle/dispatch events to the views/media player.

ziggythehamster:
Looking at that API, I'm not sure that it covers the same things. This PR gives you some more knobs to turn to customize the appearance of the waveforms (ideally, all of the constants are configurable, but we haven't needed to change everything). I'm all for abstracting that away, I'm just not entirely sure how. In an ideal world, it's controllable via CSS - but I realize you can't do that because we're talking about a canvas.

PR #126, opened 2015-07-02
Enable rendering of non-editable marker lines

ziggythehamster:
Today, in Peaks, if a marker is not editable, only the waveform highlight gets rendered. If you want the lines on either side of the marker (or if it's a point, just the line) to be rendered, that cannot be controlled. This PR adds the ability to control if we draw marker lines or not (we do not draw the handle) if the marker is not editable.

This defaults to the existing behavior; users must opt-in and choose which waveforms they want the marker lines to render on if they want this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,10 @@ var options = {
   // Zoom view adapter to use. Valid adapters are: 'animated' (default) and 'static'
   zoomAdapter: 'animated',
 
+  // Show marker lines for uneditable segments and points on the following waveforms: 'overview', 'zoomview'.
+  // Default is none of them.
+  showMarkerLinesWhenUneditable: [],
+
   // Array of initial segment objects with startTime and
   // endTime in seconds and a boolean for editable.
   // See below.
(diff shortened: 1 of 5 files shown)


ziggythehamster:

oncletom:
Hi @ziggythehamster, I am unsure to understand the actual behaviour - my head is not fresh yet ;-)

Any chance any of you to share a screenshot of the current behaviour, and a screenshot with the code of this PR?

My concern is some complexity is added in the code – it seems to do the job – but still, I wonder if it has any value to make it configurable.

ziggythehamster:
Today, if you have `editable: false` on a segment, only the waveform is highlighted - there are no bars on both sides as when `editable: true`. Points don't display at all when `editable: false`.

This PR merely makes it possible to display the bars on both sides of a segment or the single bar of a point when `editable: false`. The *handles* on the bars are not rendered.

The default behavior is exactly what Peaks has been doing already.

Also, you can configure the waveforms differently from one another. Maybe you hide the bars on the overview but show them on the zoom, or vice versa.

I can try to get you some screenshots if you'd like, though it's pretty much what you'd expect: with this option off, it looks like it does today, with this option on, it looks like editable mode just without the handles. :)

chrisn:

PR #127, opened 2015-07-02
Bugfix: Don't add waveform layer to the overview stage twice and ensure the UI layer is on top

ziggythehamster:
`createWaveform` adds the waveform layer to the stage already, but then we re-add it to the stage. I'm unsure exactly what effect this has (@johvet probably does), but it definitely seems incorrect. The UI layer also needed to be moved on top of the other layers, because the waveform was covering up the playhead and other important bits.

### DIFF
--- a/src/main/views/waveform.overview.js
+++ b/src/main/views/waveform.overview.js
@@ -44,8 +44,6 @@ define([
     that.createRefWaveform();
     that.createUi();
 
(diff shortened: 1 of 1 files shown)


ziggythehamster:

oncletom:
@ziggythehamster thanks for the reminder and fix, was away for a month and lost track of untriaged issues :-)

PR #129, opened 2015-09-22
Refactored commits for PR

lowpitch:
Have combined all my updates into one PR as none of the individual features make sense individually, and kind of need to be reviewed merged all together. 

None of the additions / changes should affect any existing implementations. All the new functionality is exposed through new configuration options, and is switched off by default. 

I've updated the example in index.html to use these new features, and have included some different audio. The existing short audio file in the demo doesn't really demonstrate the improved zoom sufficiently.

I synced my fork from master and refactored a couple of bits to allow for recent changes. Also updated to use Konva instead of Kinect. 

The work I'd been doing with Peaks didn't have any need for 'Points' and I realised that Points were broken in my updated zoom view. I updated the Points implementation so these are also now compatible - seems all good.

Summary of changes / additions:

- [ ] New `zoomMode` config option - defaults to 'stepped' which uses the previous implementation. When set to 'continuous', it will switch to using my new `ContinuousZoomView` with the new zoom functionality
- [ ] Changes to the zoom API have been made to support this - the zoom object is now a proxy, passing control to either `stepped_zoom` or `continuous_zoom` based on the config option
- [ ] New `continuousZoomCacheSize` option which gives control over how granular the pre-sampled cache of data should be. It's advised not to set this, and to let the `ContinuousZoomView` work this out for itself, but the option is there if needed
- [ ] New `showEditableOverview` option which, when set to `true` will use my updated `EditableOverviewView`. This includes the drag controls which allow you to choose the zoom level from within this window
- [ ] This `EditableOverviewView` can be expanded / contracted. The new `editableOverviewIsMaximized` config option determines whether the view is maximised when first launched. The demo page contains a button that toggles between the two states
- [ ] New `segmentStyle` option, determines whether segments will be rendered using the Rectangle shape, or the Waveform shape. Defaults to wave, but the new `ContinuousZoomView` often works better with rectangle
- [ ] As the new  `ContinuousZoomView` has an initialisation step,I've added events that report initialisation progress and completion, and have included an example of this being used in the demo

Built version of the demo page available here - http://lowpitch.com/testbed/peaks/bbc/


oncletom:
Thanks a lot!

Although it would make no sense to merge in master one by one, the ideal way would have to make multiple pull requests towards a dev branch. It would have made the review easier - at least for a reading perspective.

Is it something you can break down or it is going to need a lot of your time?

oncletom:
The demo is amazing :-) Love how smooth it is!

The segment boundaries seem not to match its filled area:

screen shot 2015-09-22 at 12 29 10

There is a weird pixel effect during on the playhead during playback:

peaks-playhead

Have you faced cases where `continuousZoomCacheSize` helped? Otherwise, better not to expose it at all.

lowpitch:
> Is it something you can break down or it is going to need a lot of your time?

You're right - sorry, very inconsiderate! I can definitely refactor into a couple of smaller PRs. Certainly the two new views can be proposed individually, and perhaps some of the minor changes to the segments class too.

> The segment boundaries seem not to match its filled area:

You're right - seems to be a problem with editable segments. My fix for 'points' was a last minute addition, and it seems not quite right for editable segments. I'll tidy that up.

> There is a weird pixel effect during on playhead during playback:

Ooh that is weird - I'll see if that can be tidied.

> Have you faced cases where continuousZoomCacheSize helped? Otherwise, better not to expose it at all.

There is one usecase that I can think of. The cache size is a function of the audio duration. For extremely long audio files, the cache might grow to an unacceptable level. So either:
a) It's useful to leave this in, and to provide better documentation for when it might be useful
or
b) Remove it, and add a hardcoded cache limit to the class itself

oncletom:
Great and don't worry, the glitches are really tiny compared to what you bring to the project :-)

As discussed over Skype, let's close this PR and create multiple ones which will be proposed against the [`1.x-dev` branch](https://github.com/bbcrd/peaks.js/tree/1.x-dev) which has just been created.

Again, amazing job, I can't wait people to get it as `1.0.0` version :-)

PR #132, opened 2015-10-08
Add a safeguard before removing segments and-/or points

ziggythehamster:


oncletom:
Hi @ziggythehamster :-) did it happen these two objects to be empty/non-existent?

chrisn:
This change no longer seems needed. Please open a new issue if there's still a problem in the latest release.

Issue #61, opened 2014-07-03
Remove points/segments option in the constructor


oncletom:
It adds unnecessary logic and can be performed as is:

``` js
Peaks.init({ … })
  .segments.add([ { startTime: 10, endTime: 12 } ])
  .points.add([ { timestamp: 11 } ]);
```

chrisn:
I'd prefer not to introduce a breaking API change at this stage. However, `Peaks.init()` does take a lot of options in addition to the ones mentioned here, and we may want to look at simplifying this at some point.

PR #137, opened 2016-06-07
add addPoint to the points api

frankbaele:
Setting points from the init options was giving function not found error because the addpoint function was never declared.


chrisn:
Thanks for reporting this bug. I've fixed it in https://github.com/bbcrd/peaks.js/commit/99f2769d889c6e9f7a02f7d4ce84a0bb051d8bdb, but see also #61.

PR #141, opened 2016-07-19
Add & remove points & segments by id

jdelStrother:
We've found it useful to be able to refer to points & segments by their ID.  How about something like this, based upon @chainlink's work in https://github.com/bbcrd/peaks.js/pull/64 ?

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -384,6 +384,7 @@ define('peaks', [
            * @param {Boolean=} pointOrPoints[].editable
            * @param {String=} pointOrPoints[].color
            * @param {String=} pointOrPoints[].labelText
+           * @param {Number=} pointOrPoints[].id
            */
           add: function (pointOrPoints) {
             var points = Array.isArray(arguments[0]) ? arguments[0] : Array.prototype.slice.call(arguments);
@@ -401,6 +402,19 @@ define('peaks', [
             points.forEach(self.waveform.points.createPoint.bind(self.waveform.points));
             self.waveform.points.render();
           },
+
+          remove: function (point) {
+            var index = self.waveform.points.remove(point);
+
+            if (index === null) {
+              throw new RangeError('Unable to find the requested point' + String(point));
+            }
+
+            self.waveform.points.render();
+
+            return self.waveform.points.points.splice(index, 1).pop();
+          },
+
           /**
            *
            * @returns {*|WaveformOverview.playheadLine.points|WaveformZoomView.zoomPlayheadLine.points|points|o.points|n.createUi.points}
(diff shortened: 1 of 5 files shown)


jdelStrother:
^ I'll look into those test failures, not sure why I didn't see those locally

jdelStrother:
Should be fixed now - I think the current failures are just sporadic ones that are fixed by #142

chrisn:
Hi Jonathan, thanks for these changes and for fixing the test failures. We'll take a look and merge soon. (Tests aren't passing locally for me on master at the moment).

jdelStrother:
Sure.   Let me know if you'd like to take a closer look at any failing tests - `npm run build && npm test -- --browsers Firefox` works ok for me (which is pretty much what Travis does), but I ran into some failures with Chrome - there's some annoying weirdness with currentTime not getting updated after seeking.

src/main/markers/waveform.segments.js
```diff
@@ -25,9 +26,9 @@ define([
       return view;
     });
 
-    var createSegmentWaveform = function (segmentId, startTime, endTime, editable, color, labelText) {
+    var createSegmentWaveform = function (id, startTime, endTime, editable, color, labelText) {
```
chrisn: I recommend keeping parameters in the same order between this and `createSegment`, and `addSegment` in main.js.

src/main.js
```diff
@@ -281,12 +281,13 @@ define('peaks', [
       get: function () {
         var self = this;
 
-        function addSegment (startTime, endTime, editable, color, labelText) {
+        function addSegment (startTime, endTime, editable, color, labelText, id) {
```
chrisn: The number of parameters is becoming unwieldy, so I'd prefer we address issue #6 before (or as part of) making this change.

jdelStrother: I can hopefully look at that sometime next week.   How about merging #142 to fix Travis?

chrisn: Done! That'd be great if you're happy to take a look at #6, otherwise I can try to find time.

jdelStrother:
I've rebased this on top of the latest master, it's a lot cleaner with the recent changes to the addPoint/addSegment method signatures.

src/main/markers/waveform.segments.js
```diff
@@ -25,10 +25,10 @@ define([
       return view;
     });
 
+    var segmentId = 0;
     var createSegmentWaveform = function (options) {
-      var segmentId = "segment" + self.segments.length;
       var segment = {
-        id: segmentId,
+        id: options.id || ('peaks.segment.' + segmentId++),
```
chrisn: A minor nitpick, but I think we should allow zero as a valid ID, and only set a default ID if undefined or null.

jdelStrother: I could - we'd need to update waveform-data as well though, which ends up replacing 0`s at https://github.com/bbcrd/waveform-data.js/blob/master/lib/core.js#L194-L195

chrisn: I see the `set_point` method in waveform-data handles this correctly, but `set_segment` is inconsistent. I'll fix and update it.

chrisn: I've just updated waveform-data.js to v1.5.2, which fixes this issue. I can't publish a new version to npm as this needs @oncletom, who's currently away, so for now I've added a git URL to package.json.

jdelStrother: Looks good, I've updated the PR with support for 0-ids.

PR #142, opened 2016-07-20
Serve media from Karma during tests

jdelStrother:
I think the second server is responsible for random test failures on Travis, we can probably make it more reliable just by serving everything through Karma.

(Unless the separate server was intentional, as an integration test for CORS..?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,5 +28,4 @@ before_script:
 - sh -e /etc/init.d/xvfb start
 - npm run build
 script:
-- npm start & sleep 2
 - npm test -- --browsers Firefox --single-run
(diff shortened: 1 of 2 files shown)


chrisn:
Many thanks!

oncletom:
@jdelStrother it does not make sense to load audio files through CORS as this is not a test case for Peaks itself — it would be the case for loading audiowaveform data files though. Which nowadays could rely on `fetch()` and expect it to be polyfilled, or using an xhr npm module, or not even fetch the data to let the user free to gather the data the way they need.

So you are right to have sent your PR and thanks @chrisn for merging it :-)

PR #143, opened 2016-08-03
Improve `addSegment` signature

jdelStrother:
It now warns on `addSegment(0, 10, …)` - a segment object should be passed instead.

eg - 
```js
  addSegment({startTime: 0, endTime: 10, …})
  addSegment([{startTime: 0, endTime: 10, …}, {startTime: 10, endTime: 20, …}])
```

---

This is something of a work in progress - I want to change `points.add` to follow the same signature, but thought I should get feedback on whether this is going in the right direction first.   In particular, I think it's a little weird that `segments.addSegment` now takes a segment-object, but then breaks it back out to spread-arguments for `waveform.segments.createSegment`.   WDYT to just passing the segment-object directly to createSegment ?

(Previously: #141 & #6)

### DIFF
--- a/test/unit/api-points-spec.js
+++ b/test/unit/api-points-spec.js
@@ -29,16 +29,16 @@
       });
 
       it("should return any added point", function () {
-        p.points.add(10, true, "#ff0000", "A point");
-        p.points.add(12, true, "#ff0000", "Another point");
+        p.points.add({timestamp: 10});
+        p.points.add({timestamp: 12});
         expect(p.points.getPoints()).to.have.length.of(2);
       });
     });
 
 
     describe("add", function () {
-      it("should create a segment with the mandatory values as arguments", function () {
-        p.points.add(10, true, "#ff0000", "A point");
+      it("should create a segment from the supplied object", function () {
+        p.points.add({ timestamp: 10, editable: true, color: "#ff0000", labelText: "A point"});
 
         var point = p.points.getPoints()[0];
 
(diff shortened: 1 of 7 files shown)


chrisn:
Your changes look great, thank you! You're right, `createSegment` should take a segment object too, and this would simplify the code. If you're happy to update the points interface, that would be very welcome!

jdelStrother:
Is backwards compatibility needed on createSegment, or can we just replace
all calls of createSegment with an object?

chrisn:
I would say that `createSegment` is an internal method and not part of the public API, so no need to preserve backwards compatibility. Application code should be using `instance.segments.add`.

jdelStrother:
I swear I added a comment here earlier, not sure where it got to.  Apologies if this crops up twice:

---

`segments.add` and `points.add` now accept an object rather than spread-arguments.   The latter now gives a deprecation warning.

Internally, createSegment etc just accept objects, and will throw an exception if you try to use the old `createSegment(startTime, endTime...)` signature.

I've got a couple of commits on top of this for adding/removing points & segments by their ID, which falls quite nicely out of the new method signature changes (https://github.com/jdelStrother/peaks.js/compare/method-signatures...jdelStrother:edit-by-id).   Would you like them as a separate PR or shall I push them to this branch?

test/unit/api-points-spec.js
```diff
@@ -57,48 +57,42 @@
...
       it("should throw an exception if timestamp argument is null", function () {
         expect(function () {
-          p.points.add(null);
+          p.points.add({timestamp: null});
         }).to.throw(TypeError);
       });
 
       it("should throw an exception if the timestamp argument is missing", function () {
         expect(function () {
-          p.points.add(NaN);
```
chrisn: We should keep the `NaN`tests, as there's code that checks for these.

jdelStrother: Yep, good catch

chrisn:
I'd prefer to have the ID changes in a separate pull request, if that's OK. Thanks again for your efforts!

jdelStrother:
Ok, I've restored that NaN-timestamp test.

chrisn:
All looks good, so I'll merge this.

PR #144, opened 2016-08-05
Use Konva.FastLayer for drawing waveforms

jdelStrother:
FastLayer doesn’t support, eg, hit-testing, which makes it way faster when rendering large waveforms.

### DIFF
--- a/src/main/views/waveform.overview.js
+++ b/src/main/views/waveform.overview.js
@@ -29,7 +29,8 @@ define([
       height: that.height
     });
 
-    that.waveformLayer = new Konva.Layer();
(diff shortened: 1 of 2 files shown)


chrisn:
Interesting, do you have any advice on how to test the performance improvement?

jdelStrother:
I was mostly manually inspecting it using Chrome's Timeline tab, especially while zooming in & out of hour-long mp3s.  Time was split roughly equally between `drawScene` and `drawHit` - the latter disappears with FastLayer.

chrisn:
Yes, I see the difference. Nice change, thank you!

PR #150, opened 2016-08-19
Peaks.destroy

jdelStrother:
This branch adds a Peaks.destroy method, used to clean up event listeners on the window and the media element.   WDYT?

### DIFF
--- a/src/main/views/waveform.overview.js
+++ b/src/main/views/waveform.overview.js
@@ -231,5 +231,10 @@ define([
     this.uiLayer.draw();
   };
 
+  WaveformOverview.prototype.destroy = function () {
+    this.stage.destroy();
+    this.stage = null;
+  };
+
   return WaveformOverview;
 });
--- a/src/main/views/waveform.zoomview.js
+++ b/src/main/views/waveform.zoomview.js
@@ -400,5 +400,10 @@ define([
     this.updateZoomWaveform(this.frameOffset);
   };
 
+  WaveformZoomView.prototype.destroy = function () {
+    this.stage.destroy();
+    this.stage = null;
+  };
+
   return WaveformZoomView;
 });
--- a/src/main.js
+++ b/src/main.js
@@ -567,5 +567,14 @@ define('peaks', [
     }
   });
 
+  /**
+   * Clean up event listeners when you're no longer using a Peaks instance
+   */
+  Peaks.prototype.destroy = function() {
+    this.removeAllListeners();
+    this.waveform.destroy();
+    this.player.destroy();
+  };
+
   return Peaks;
 });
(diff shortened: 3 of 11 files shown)


chrisn:
Note: see also https://github.com/bbcrd/peaks.js/issues/69.

src/main/views/waveform.overview.js
```diff
@@ -231,5 +231,10 @@ define([
     this.uiLayer.draw();
   };
 
+  WaveformOverview.prototype.destroy = function () {
+    this.stage.destroy();
+    this.state = null;
```
chrisn: Do you mean `this.stage`?

src/main/views/waveform.zoomview.js
```diff
@@ -400,5 +400,10 @@ define([
     this.updateZoomWaveform(this.frameOffset);
   };
 
+  WaveformZoomView.prototype.destroy = function () {
+    this.stage.destroy();
+    this.state = null;
```
chrisn: `this.stage`?

src/main/views/waveform.overview.js
jdelStrother: Almost certainly, sorry.   I'll push a fix.

chrisn:
Thanks again for your contributions. Could you tell us about the problems you're seeing that require this change, and any testing you've done to show that this PR fixes them (eg, memory profiling?).

jdelStrother:
We're using this in a single-page React app, and we create & discard multiple peaks instances through the duration of the page.   The event listeners on the discarded ones would stick around though, so we'd quite often see the error "options.width should be a strictly positive integer value" during resizing the window, for example.   That error was the main impetus, rather than having problems with memory usage.

chrisn:
Was the error repeatable? It'd be nice to have a test case around it.

jdelStrother:
I’ll see what I can do.

jdelStrother:
I managed to come up with a test for it... it's not pretty, but see what you think.   https://github.com/bbcrd/peaks.js/pull/150/files#diff-d83a089f2fdd993eaf02f77a31969e49R92

chrisn:
Looks good to me, and tests OK locally in Chrome and Firefox. Is this ready to merge from your point of view?

jdelStrother:
Yes please!

oncletom:
Small question @jdelStrother: why do you use your own `listeners` array and `_addMediaListener` instead of relying on the [`EventEmitter2.removeAllListeners`](https://github.com/asyncly/EventEmitter2#emitterremovealllistenersevent) method – Peaks class extends EventEmitter2?

jdelStrother:
Can you use eventemitter for binding native events?

oncletom:
@jdelStrother hm, no and looking at the code, I don't think it will remove much of what you already wrote. It would be useful just a way to proxy the native media element events.

This is very good like it is at the moment, thanks for that :-)

PR #151, opened 2016-08-19
Report XHR errors

jdelStrother:
If the XHR fails for some reason (perhaps because of [annoying cross-origin-redirect bugs](https://bugs.webkit.org/show_bug.cgi?id=57600)), the error is quietly swallowed.   How about reporting it via `peaks.emit('error')` ?

### DIFF
--- a/src/main/waveform/waveform.core.js
+++ b/src/main/waveform/waveform.core.js
@@ -107,6 +107,10 @@ define([
           }
         };
 
+        xhr.onerror = function() {
+          peaks.emit('error', new Error('XHR Failed'));
+        };
+
         xhr.send();
       },
 


chrisn:
This looks useful, thank you. Is there a way to test this? (I had a quick look at Sinon's `useFakeXMLHttpRequest` but this doesn't seem to support triggering the `onerror`.)

jdelStrother:
You could do so in a somewhat artificial way by trying to load a URI like `file:///not-a-file`.

chrisn:
I've tested this with a `file://` URL, but this causes the error to be emitted before the caller has had a chance to add an error handler with `instance.on('error')`. I'll merge and add a fix for this. Thanks!

chrisn:
We need a better test case for this change, as a CORS error causes the `xhr.send()` call to throw an exception before the event is emitted (in Firefox at least).

oncletom:
@chrisn I wonder if it would be better to:

1. delegate the xhr bit to either a well tested Node module
2. to any other kind of library (and the user provides the results)
3. refactor this as a more solidly tested module used internally by Peaks

chrisn:
I think I'm OK with it being internal to Peaks, and there's probably a good HTTP request module we can use. The motivation behind this change is good though - applications should get visibility of the errors so they can act appropriately. I was thinking something like:

```javascript
p = Peaks.init({ ... });
p.on('error', errorCallback);
p.load('https://url-of-data');
```

which would also allow applications to change the waveform being displayed. But, this introduces more complexity in implementation.

jdelStrother:
If you're going to use an external module, maybe we should just rely on `fetch` being available, and users will have to polyfill it themselves if they're not happy with the browser support?

Or if you want to keep it internal, maybe just defer the `xhr.send()` (`setTimeout(function(){xhr.send})`) to allow users chance to add the error handler?

oncletom:
There is definitely a benefit of separating the waveform container drawing, data fetching and data drawing – especially to speed up the load on the client side.

Using `fetch` and directing to various ways of polyfilling it (via polyfill.io or [github/fetch](https://github.com/github/fetch)). Peaks would benefit of relying on promises, especially for error handling and it would help reporting/exposing them via events.

PR #157, opened 2016-09-02
Register mouse up/move events to the window rather than the waveform stages

### LINKED ISSUE #149: Waveform scrolling continues after mouseup
Dragging to scroll the waveform in the zoom view can result in the waveform continuing to scroll when the mouse is moved, after the user has released the mouse button. The problem occurs in Chrome and Firefox on Ubuntu (possibly also other browsers and OSs).

To see the problem, open the Peaks.js Demo Page (index.html).

1. Click on the zoom view and hold the mouse button down
2. Move the cursor outside the zoom view
3. Release the mouse button
4. Move the cursor within the zoom view
5. Notice that the zoom view still scrolls even though the mouse button has been released

This problem is caused (in `waveform.zoomview.js`, and likely also in `waveform.overview.js`) when the user releases the mouse button outside the view container, as the view never receives the `mouseup` event.

jdelStrother:
Fixes #149 

WDYT?

### DIFF
--- a/README.md
+++ b/README.md
@@ -484,7 +484,8 @@ Peaks.js emits events to enable you to extend its behaviour according to your ne
 ---------------------------|----------------
 `segments.ready`           | N/A
 `points.ready`             | N/A
-`segment.dragged`          | `Segment segment`
+`segments.dragged`         | `Segment segment`
+`points.dragged`           | `Point point`
 
 # Building Peaks.js
 
--- a/src/main/markers/waveform.points.js
+++ b/src/main/markers/waveform.points.js
@@ -141,8 +141,9 @@ define([
       point.timestamp = thisPoint.view.data.time(inOffset);
     }
 
+    this.peaks.emit('points.dragged', point);
     this.updatePoint(point);
-    self.render();
+    this.render();
   };
 
   WaveformPoints.prototype.init = function() {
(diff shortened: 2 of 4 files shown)


chrisn:
I've just tested this, and it certainly fixes the scrolling after mouse-up problem. But I notice that when I move the mouse in and out of the waveform view while holding the mouse button down, the waveform position jumps. I wonder if there's a way to avoid this, so the waveform position doesn't change (which was the previous behaviour).

jdelStrother:
Ugh, yeah - looks like we shouldn't be using layerX there.   Will get it fixed.

jdelStrother:
OK, updated.  I wasn't entirely sure what this code was trying to do:
```js
dX = event.evt.layerX > x ? x - event.evt.layerX : (x - event.evt.layerX) * 1;
```
- should that `1` have been a `-1` or something?   I ended up just replacing it with this - 

```js
dX = x - event.clientX
```

chrisn:
Yeah, that original code certainly isn't right. I guess the intent might have been to ensure `dX` is always positive? Your change works fine though, looks good to merge. Thanks!

One other thing I've noticed, unrelated to this change, is that dragging in the waveform overview doesn't update the playhead position, so you can end up with the playheads in the two views in different positions. It doesn't look to be a regression, as I've just tested a few older versions. I should raise this as a new issue.

PR #161, opened 2016-10-27
More reliable clicking behaviour, doesn’t turn seek click off with ve…

Develliot:
…rtical and sub pixel mouse movements.

### DIFF
--- a/src/main/views/waveform.zoomview.js
+++ b/src/main/views/waveform.zoomview.js
@@ -70,52 +70,56 @@ define([
     // INTERACTION ===============================================
 
     self.stage.on('mousedown', function(event) {
+      var x, dX, p;
+
+      // enable drag if necessary
+      function mouseMove(event) {
+        var pixelsDifference = Math.abs(event.layerX - x);
+
+        if (pixelsDifference < 1) {
+          return;
+        }
+        peaks.seeking = false;
+
+        dX = x - event.layerX;
+        x = event.layerX;
+
+        p = self.frameOffset + dX;
+
+        if (p < 0) {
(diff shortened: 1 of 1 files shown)


chrisn:
Sorry for the delay. Fixing the click behaviour on vertical movement is a nice addition, thank you. On what browsers / platforms do you see sub-pixel mouse movements?

Develliot:
No worries

Chrome on Mac using a 4k screen.

Develliot:
Sometimes I had to double click to move the play head

src/main/views/waveform.zoomview.js
```diff
@@ -80,7 +80,10 @@ define([
 
           // enable drag if necessary
           var mouseMove = function(event) {
-            peaks.seeking = false;
+			var pixelsDifference = Math.abs(event.layerX - x);
```
chrisn: I don't have access to either a Mac or a 4K display. Can `event.layerX` have a non-integer value?

Develliot: I don't believe it can be anything other than an integer as it would always represent a coordinate value as a number of pixels

Develliot: Scratch that I was looking at looking at MDN documentation but there is the possibility that this is not the case with the way chrome deals with high resolution screens

Develliot: After further testing it is possible to have sub pixel values with css but not with JS move events, they are full integers for now.

chrisn:
Great! Thank you for the updates and for investigating the sub pixel question.

PR #166, opened 2016-11-29
Expose updatePoints & updateSegments

Develliot:
Useful if you have another datasource modifying, eg, point timestamps and you need peaks to redraw

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -381,6 +381,14 @@ define('peaks', [
 
           removeAll: function() {
             return self.waveform.segments.removeAll();
+          },
+
+          /**
+           * Redraw segments
+           *
+           */
+          updateSegments: function() {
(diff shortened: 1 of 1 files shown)


chrisn:
For the use case you mention, I'd prefer to see a `point.setTime` method or similar. All manipulation should go through the points and segments APIs, rather than modifying the underlying data directly, to preserve encapsulation.

Refactoring the existing points and segments implementations is on our list of things to do, so I wouldn't want applications to depend on existing implementation details.

chrisn:
Closing, as the [Points API](https://github.com/bbc/peaks.js/blob/master/README.md#points-api) and [Segments API](https://github.com/bbc/peaks.js/blob/master/README.md#segments-api) should provide everything you need. Please raise a new issue if these don't work for your use case.

PR #167, opened 2016-11-29
Removal of audioContext creation, now passed in in the config object

dodds-cc:
The audioContext is now passed into the Peaks instance and passed to the builder.
Do you think this is the way to go about it?

relies on bbc/waveform-data.js#37 and bbc/waveform-data.js#38


dodds-cc:
Just trying to work out why these tests are failing on build here but all passing when building locally

oncletom:
I will check this out tomorrow @dodds-cc :-)

dodds-cc:
@oncletom Thanks, its really quite tricky to debug. I was thinking of rolling my firefox back to the version the browserstack is using but maybe I'm missing something obvious somewhere..

PR #171, opened 2016-12-07
1.x.dev segment edit

alanbacon:
Editable segments now show and move correctly. There was an issue when dragging that the new pixel position was derived by converting from the dragged pixel position into a time and then back into a pixel position. This caused a precision/rounding error where the derived pixel position of both handles was different from the pixel position of where the handle was dragged to. The effect was both handles of the segment would move uncontrollably when dragging either handle. When dragging the pixel position is now used as the ground truth, instead of the intermediate derived time value


chrisn:
Closing, as draggable segments is now in v3.0.0 beta

PR #173, opened 2016-12-10
Remove jshint and comply to eslint rules

oncletom:
No warnings and no errors anymore


src/main/waveform/waveform.core.js
```diff
@@ -66,7 +66,7 @@
    * data) or 'json'
    * @param {HTMLMediaElement} options.mediaElement
    *
-   * @see Refer to the <a href="https://github.com/bbc/audiowaveform/blob/master/doc/DataFormat.md">data format documentation</a>
+   * @see Refer to the <a href="../../../doc/DataFormat.md">data format documentation</a>
```
chrisn: This link is to a [doc](https://github.com/bbc/audiowaveform/blob/master) in the audiowaveform repo.

oncletom: Oh my bad! Because I was switching between the two repo I thought it was written as an absolute link and could be relative instead. Will rectify that :-)

package.json
```diff
@@ -42,14 +42,14 @@
...
   "scripts": {
+    "prebuild": "npm run lint",
     "build": "npm run build:main && npm run build:standalone",
     "build:main": "browserify -e ./src/main.js -t deamdify | derequire - > index.js",
     "build:standalone": "browserify -d -e ./src/main.js -t deamdify -s peaks | exorcist peaks.js.map | derequire - > peaks.js",
     "doc": "jsdoc --private --destination docs --recurse src",
     "lint": "eslint src/**/*.js test/**/*.js",
     "pretest": "npm run build",
     "test": "./node_modules/karma/bin/karma start",
-    "posttest": "npm run lint",
```
oncletom: @chrisn it is actually better to me to consider the linting as being a pre or post test task. Build can in fact happen without a harm if eslint rules do not comply.

Having it in `prebuild` is an implicit assumption the `build` step happens at `test` time – which can change if the browserify karma plugin is enabled for example (no need to build anymore).

What do you think?

chrisn: IMHO you want to run the lint as early as possible, with the intent of exposing errors in the code.

PR #175, opened 2016-12-20
Upgrade development environment workflow

thom4parisot:
The goal of this pull request is to speed up operations and to clear up around few concepts:

- *testing* is about reaching quality goals
- *building* is about producing artifacts
- *publishing* is about testing then building then pushing — and it should be done by the CI only, after tagging a git commit

# Development

`npm test` only has to run tests now, and modules can be required with `require()` + browserify directly.
It will enable atomic testing of individual peaks modules.

`npm run watch` is the central command to run test continuously, start a static files server and run browserify continuously too.

# Demo 

`npm start` will build artifacts-to-be-published and expose them via static files server.


chrisn:

PR #176, opened 2016-12-20
Reinstate bower dependencies and provide an alternative demo loading method

oncletom:
The only drawback is because `waveform-data` does not publish its `dist` folder anymore, there is no "non-build" way to load all bower components.

In the demo, I had to load waveform-data from a CDN which does the browserify build. Although it would not work with the webaudio loader :-(

Two possible solutions:

1. publish again the `dist` folder in git for waveform-data (but I find it ugly and not much compatible with a good CI flow)
1. we don't mind and drop bower in a later version of peaks (it will be a limitation for ES2015 modules anyway, and bower has no much future)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ requirejs.config({
   paths: {
     peaks: 'bower_components/peaks.js/src/main',
     EventEmitter: 'bower_components/eventemitter2/lib/eventemitter2',
-    Konva: 'bower_components/konvajs/konva',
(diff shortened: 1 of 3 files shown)


chrisn:
I'll close this, as Bower support was removed in https://github.com/bbc/peaks.js/commit/b0e2d95fbd7411b8cb1ad906210c74f7afeac957.

Issue #59, opened 2014-07-02
Toggle a view on/off


oncletom:

chrisn:
See also #53. The refactoring done in the v0.9.0 release is a step towards enabling configurable views.

chrisn:
The [Views API](https://github.com/bbc/peaks.js/blob/master/README.md#views-api) contains methods that allow applications to toggle the waveform views. See [demo/index.html](https://github.com/bbc/peaks.js/blob/master/demo/index.html) for an example.

Issue #46, opened 2014-05-26
Visualizing all audio and zooming on the same canvas


katcipis:
Is there a way to show only one waveform and zoom on it instead of always having two ? I have some space constraints and it would be awesome to use only one waveform (one canvas) to zoom in/out and show the entire file.

All the examples always shows two, so i don’t know if it is possible.

If it is not possible right now, would it be an welcomed addition ? :-)

chrisn:
Release 0.10.0 allows you to create a single waveform view. Please see https://github.com/bbc/peaks.js/blob/master/demo/zoomable-waveform.html for an example.

PR #177, opened 2016-12-20
Merge ZoomView and Overview into a single WaveformView object

thom4parisot:
The main benefit is to have a single codebase with features to work out what a waveform is.

It will allow in the future to have as many waveform views as necessary:

- refs #53 
- refs #2 
- refs #59 
- refs #46


chrisn:

PR #179, opened 2016-12-21
Replace `Peaks.init` with `new Peaks()` constructor form

oncletom:
Add deprecation warning as well.


chrisn:
The code changes look OK, but any idea what's causing the test failures?

chrisn:
See #186.

PR #180, opened 2016-12-22
[WIP] Unify Points and Segments API as Marker API

### LINKED ISSUE #87: Markers API

thom4parisot:
closes #87


chrisn:

PR #181, opened 2016-12-29
Distinguish between played and unplayed portions of the audio

jtolla:
screen shot 2016-12-28 at 9 14 49 pm

This adds an optional semi-transparent overlay on top of the played portion of the audio. 

Ideally, I think this would also overlay segments and points, but since the playhead and the rest of the uiLayer are currently underneath the S+P's, it could make sense to bring the whole uiLayer to the forefront (along with this overlay) as part of another PR.

If you think bringing the uiLayer out from behind the S+P's is something that makes sense, I'd be happy to give it a go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,9 @@ var options = {
   // Zoom view adapter to use. Valid adapters are: 'animated' (default) and 'static'
   zoomAdapter: 'animated',
 
+  // Fade the played portions of the waveforms
+  playedOverlay: true,
+
   // Array of initial segment objects with startTime and
   // endTime in seconds and a boolean for editable.
   // See below.
--- a/src/main.js
+++ b/src/main.js
@@ -216,7 +216,13 @@ define('peaks', [
       /**
        * Use animation on zoom
        */
-      zoomAdapter: 'animated'
+      zoomAdapter: 'animated',
+
+      /**
+       * A semi-transparent waveform overlay, used to distinguish between
+       * played and unplayed portions of the audio.
+       */
+      playedOverlay: true
     };
 
     /**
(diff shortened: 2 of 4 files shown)


chrisn:
Thanks for the contribution, looks good. A couple of comments, though:
* This only handles the common case of playing the audio from the beginning. What should happen if the user seeks forward? Should we keep track of each separate region that's been played?
* I agree with you about ordering of the layers.

jtolla:
Specifically overlaying _only_ played portions sounds like an interesting and useful feature, and one that I'm going to plan to work on. 

That said, I think there's value in merging as is for now.  The way it's currently set up seems to be inline with what Sound Cloud and others do ([example](https://soundcloud.com/gimletmedia/undone-preview)). It's more of a "this is in the past" overlay, than something that represents played content.

I think your first question was leading into the second, but answering it in isolation, you should be able to drag/seek and the overlay should update accordingly.

chrisn:
I just did a quick experiment, and simply bringing the uiLayer to the top makes it impossible to drag the point / segment handles. I'd prefer to see a solution to this before merging this PR.

amrfaisal:
+1 For merging this pull request, as this feature is really useful.

ffxsam:
This would be super cool. Also, a way to preload those played segments on the `Peaks.init()` call, e.g.:

```js
Peaks.init({
  ...options,
  playedRegions: [
    { startTime: 23, endTime: 38 },
    { startTime: 64, endTime: 213 }
  ]
});
```

chrisn:
I was seeing this as a simple feature where the audio region before the playhead has one colour, and the remaining audio has a different colour. Allowing arbitrary played/unplayed regions adds quite some complexity.

chrisn:
This is now implemented, see [view.setPlayedWaveformColor()](https://github.com/bbc/peaks.js#viewsetplayedwaveformcolorcolor).

PR #192, opened 2017-06-16
Add play segment function and update demo to showcase it

craigharvi3:
### Changes

* Added `Player.playSegment(segment)` function.
* Update demo to showcase new function

### DIFF
--- a/index.html
+++ b/index.html
@@ -113,6 +113,7 @@
                 <th>Label</th>
                 <th>Start time</th>
                 <th>End time</th>
+                <th></th>
               </tr>
             </thead>
             <tbody>
@@ -162,6 +163,7 @@
                 '<td>' + ((segments[i].labelText) ? segments[i].labelText : 'N/A' ) + '</td>' +
                 '<td>' + segments[i].startTime + '</td>' +
                 '<td>' + segments[i].endTime + '</td>' +
+                '<td>' + '<a href="#' + segments[i].id + '" data-action="play-segment" data-id="' + segments[i].id + '">Play segment</a>' + '</td>' +
               '</tr>'
               html += row;
             }
@@ -178,7 +180,6 @@
             var html = '';
 
             for( var i=0; i<points.length; i++ ) {
-              console.log(points[i]);
               var row = '<tr>' +
                 '<td>' + points[i].id + '</td>' +
                 '<td>' + points[i].timestamp + '</td>' +
--- a/index.html
+++ b/index.html
@@ -3,7 +3,8 @@
   <head>
     <title>Peaks.js Demo Page</title>
     <style>
-      #titles {
+
+      body {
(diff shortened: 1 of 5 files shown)


craigharvi3:
@chrisn Feedback is most welcome! Cheers.

src/main/markers/waveform.segments.js
```diff
@@ -330,6 +330,16 @@ define([
     return this.segments;
   };
 
+  WaveformSegments.prototype.getSegment = function(id) {
+    for ( var i = 0; i < this.segments.length; i++ ) {
```
chrisn: Style nitpick: although not in our eslint config, this project doesn't use spaces inside parentheses (we should add `"space-in-parens": ["error", "never"]` to `.eslintrc.js`).

src/main/player/player.js
```diff
@@ -158,5 +160,30 @@ define(['peaks/waveform/waveform.utils'], function(Utils) {
...
+   *
+   * @param {Segment} segment
+   */
+  Player.prototype.playSegment = function(segment) {
+    var self = this;
+
+    clearTimeout(self._interval);
+
+    // Set audio time to segment start time
+    self.peaks.time.setCurrentTime( segment.startTime );
```
chrisn: No need to call via `self.peaks` here, prefer `self.seekBySeconds`.

src/main/player/player.js
```diff
@@ -158,5 +160,30 @@ define(['peaks/waveform/waveform.utils'], function(Utils) {
...
 
+  /**
+   * Plays the segment passed in
+   *
+   * @param {Segment} segment
+   */
+  Player.prototype.playSegment = function(segment) {
+    var self = this;
+
+    clearTimeout(self._interval);
```
chrisn: Set `self._interval = null` after calling `clearTimeout()` (and at line 182).

chrisn: Also call `clearTimeout()` to tidy up in `Player.destroy()`.

index.html
```diff
@@ -80,6 +80,10 @@
         <button data-action="log-data">Log segments/points</button>
       </div>
 
+      <ul id="segments">
```
chrisn: The layout of the demo page needs some work, to align the segments with the other page content. I'm thinking that instead of showing each segment as it's added we should change the `log segments/points` button to show the segments (and points) in the page, rather than in the console.

craigharvi3: @chrisn What info should be displayed for the segments and points when outputted to the page?

chrisn: What I would do is output two tables, one for segments, the other for points. The segments table would show the `startTime`, `endTime`, `id`, and `labelText` attributes, and the points table would show `timestamp`, `id`, and `labelText`.

craigharvi3:
Ready for feedback @chrisn

chrisn:
Looks good, thanks!

PR #202, opened 2017-11-08
Keep playhead layer in sync with timeupdate events

jdelStrother:
Currently, when zoomview receives a timeupdate event from the player, it calls `_playheadLayer.syncPlayhead`.  However, if a play animation is present (from `PlayheadLayer.playFrom`), that will immediately call syncPlayhead again with the time value based on when playFrom was last called.

This commit calls playFrom on each timeupdate event, which eliminates drift and ensures that the playhead is in sync with any changes made to the audio node (eg if the audio node is playing and you change audio.currentTime externally to peaks.js).

WDYT?

### DIFF
--- a/index.html
+++ b/index.html
@@ -43,6 +43,14 @@
         cursor: pointer;
       }
 
+      #seek-time {
+        width: 4em;
+      }
+
+      #controls {
+        float: right;
+      }
+
       .log {
         margin: 0 auto 24px auto;
         width: 1000px;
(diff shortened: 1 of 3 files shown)


chrisn:
Thanks for this. Looks like I may need to rethink this refactoring... Do you have an example page you could share that highlights the problem?

jdelStrother:
https://codepen.io/jdelStrother/pen/QOaVax?editors=1010

Try starting playback, then hit the "Jump to t=10" button.  The audio output jumps, as does the marker on the zoomed-out view, but the marker on the zoomed-in view continues animating on its original trajectory.

You can also see the drift problem simply by starting playback, then pausing after a few seconds.   If you look closely at the marker on the zoomed-in view, the marker jumps backwards as you pause it, because the animation slowly gets out of sync with the audio's actual currentTime, and then the marker moves back to the correct currentTime on pause.

chrisn:
LGTM. I'll merge and prepare a new release if you're happy with my additional changes. Thanks.

jdelStrother:
Yes please, that’d be awesome.

PR #204, opened 2017-11-27
Fix the overview playhead not keeping in sync following a seek

jdelStrother:
Sorry - one more fix for the playhead animations. 

If you play and then seek by clicking in the `zoomview`, the playhead in the `overview` view jumps around.  It's keeping the original play animation but also receiving timeupdate events for a different time.  This commit ensures the two are in sync.

It wasn't visible in my original demo page because it was a long audio file, so the overview waveform was extremely zoomed out.  It's more obvious with a shorter clip here - https://codepen.io/jdelStrother/pen/QOaVax?editors=1010.

[WaveformOverview & WaveformZoomview are building up quite a bit of duplication.  How do you feel about extracting the duplication to a shared set of functions?]

### DIFF
--- a/src/main/views/waveform.overview.js
+++ b/src/main/views/waveform.overview.js
@@ -90,11 +90,7 @@ define([
 
         var time = self.pixelsToTime(mousePosX);
 
-        self._playheadLayer.syncPlayhead(mousePosX);
-
-        if (self._playing) {
-          self._playheadLayer.playFrom(time);
-        }
+        self.updatePlayheadTime(time);
 
         self.peaks.emit('user_seek', time);
       },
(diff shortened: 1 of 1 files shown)


chrisn:
I wonder how feasible it is to write some unit tests around this. I'll give this some thought, but suggestions welcome :-)

> [WaveformOverview & WaveformZoomview are building up quite a bit of duplication. How do you feel about extracting the duplication to a shared set of functions?]

I have unpublished work in progress to unify these. I'll post an issue on this soon...

jdelStrother:
I did consider unit tests and decided it was too hard 😞

chrisn:
The changes look good. Thank you!

Issue #205, opened 2017-11-28
Playhead jumps shortly after starting playback


chrisn:
Following #199, #202, and #204, one small issue with the playhead marker remains: The playhead marker sometimes jumps backwards shortly after starting playback. This can be seen in zoomable waveform view in the Peaks.js demo page.

cc @jdelStrother if you have any ideas?

jdelStrother:
Confirmed.  Will see what I can do

chrisn:
Thanks! I'll hold off on publishing 0.9.5 while we investigate.

jdelStrother:
Yeah....not entirely sure of the best way of dealing with this.  If you look at the audio node's `currentTime` when the play event is received, and compare it to `currentTime` when subsequent timeupdate events arrive, it's consistently around 60ms behind the elapsed wall clock time, which I think is what's causing the jump.  Chrome sends timeupdate events every 250ms, so I _think_ that 250ms after you press play, the playhead jumps backwards 60ms.

I've experimented with a couple of options...

* don't start the play animation when the play event is received - instead, start it on timeupdate.   It looks less broken because the playhead doesn't animate forwards then jump backwards - it just jumps forwards - but it does feel pretty laggy.

* Ditch the playhead's `Konva.Animation` entirely.   Instead, during playback, continually fire requestAnimationFrame, read audio.currentTime, and update the playhead's position from that.  I think this is the most promising - there is a tiny jump in the playhead a few milliseconds after playback starts, but it's much less noticeable, and is otherwise pretty smooth.

Any thoughts?

chrisn:
Resolved via #207.

PR #206, opened 2017-11-29
Replace playhead Konva animations with requestAnimationFrame

jdelStrother:
As discussed in #205, HTML5 audio emits `timeupdate` events at a very low frequency - around 250ms in Chrome.
This PR uses requestAnimationFrame to continually fire `player_time_update` events during playback.  This removes the need for a custom animation in the playhead layer, and makes any jumps in currentTime much less noticeable.

WDYT?

### DIFF
--- a/src/main/player/player.js
+++ b/src/main/player/player.js
@@ -29,6 +29,7 @@ define(['peaks/waveform/waveform.utils'], function(Utils) {
     self._listeners = [];
     self._mediaElement = mediaElement;
     self._duration = self.getDuration();
+    self._isPlaying = false;
 
     if (self._mediaElement.readyState === 4) {
       self._peaks.emit('player_load', self);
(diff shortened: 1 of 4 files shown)


chrisn:
Thank you! You're right that `timeupdate` events are only generated infrequently, so your approach of polling the `currentTime` is good. As we don't have other use cases for more frequent `timedate` events (AFAIK), I'd prefer to limit the changes to the view related code, if possible, rather than modifying the `Player` object. 

Also, I had implemented logic to switch off the animation at zoom levels where using `timeupdate` events alone is good enough, as `requestAnimationFrame` is quite CPU intensive, so I'd prefer not to remove this code.

So I wonder if a simpler solution it this change to `PlayheadLayer.playFrom`? Or maybe i'm missing something?

```javascript
     self._playheadLineAnimation = new Konva.Animation(function(frame) {
-      // Elapsed time since animation started (seconds).
-      var elapsed = frame.time / 1000;
-
-      var playheadPosition = self._view.timeToPixels(startTime + elapsed);
+      var time = self._peaks.player.getCurrentTime();
+      var playheadPosition = self._view.timeToPixels(time);
     
       if (playheadPosition !== lastPlayheadPosition) {
         ... etc.
```

jdelStrother:
Sure.  I rather liked taking all of the isPlaying logic out of the view, but that approach is good too.

chrisn:
You're right, the `isPlaying` flag is better placed inside the `Player` object. Feel free to add that to your #207 PR, or let me know if you'd prefer me to do it.

chrisn:
Closing, as superseded by #207.

PR #207, opened 2017-12-01
Rely on player.currentTime for syncing the playhead animation

jdelStrother:
An alternative approach to #206

### DIFF
--- a/src/main/player/player.js
+++ b/src/main/player/player.js
@@ -29,6 +29,7 @@ define(['peaks/waveform/waveform.utils'], function(Utils) {
     self._listeners = [];
     self._mediaElement = mediaElement;
     self._duration = self.getDuration();
+    self._isPlaying = false;
 
     if (self._mediaElement.readyState === 4) {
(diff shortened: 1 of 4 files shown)


jdelStrother:
I wonder if the pixelsPerSecond limit (https://github.com/bbc/peaks.js/blob/5855ae58f7645252941de957fda9f80df4923159/src/main/views/playhead-layer.js#L57) should be lowered.  On the demo page sample, the overview waveform doesn't have animations enabled, and the playhead usually jumps 2 pixels at a time.

jdelStrother:
👍

chrisn:
I made a change to ensure that the playhead time label updates when zoomed out to levels where the `Konva.Animation` isn't used. Updates are driven by the media element's `timeupdate` event, so occur much less frequently than using the animation loop. I think this is OK, and better than only updating the label when the playhead moves pixel position, which can be only every few seconds, depending on zoom level. If you're happy with this, I think we can merge.

jdelStrother:
Thanks, looks good to me

PR #209, opened 2018-01-16
Add cdnjs installation in README.md

extend1994:
Same as the heading.


chrisn:
Thanks for this :-) Happy to see the library available through cdnjs, even though we didn't submit it ourselves.

For reference here are the issues where it was added: https://github.com/cdnjs/cdnjs/issues/11016, https://github.com/cdnjs/cdnjs/pull/11138.

PR #213, opened 2018-01-23
Adds an option to set xhr.withCredentials when instantiating peaks

bennypowers:
Straightforward PR which allows users to send credentials when fetching files

### DIFF
--- a/src/main/waveform/waveform.core.js
+++ b/src/main/waveform/waveform.core.js
@@ -273,6 +273,10 @@
       }, 0);
     };
 
+    if (isXhr2 && this.peaks.withCredentials) {
+      xhr.withCredentials = true;
+    }
+
     return xhr;
   };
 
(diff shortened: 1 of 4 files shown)


chrisn:
This is a useful addition, thank you! If feasible, would you mind adding unit tests to this PR, please?

src/main/waveform/waveform.core.js
```diff
@@ -273,6 +273,10 @@
       }, 0);
     };
 
+    if (isXhr2 && this.peaks.withCredentials) {
```
chrisn: The `withCredentials` flag should be part of the `options` object passed to `Peaks.init()`, rather than being set on the `peaks` object itself. Also, a default value (`false`) should be added to the `options` object in [main.js](https://github.com/bbc/peaks.js/blob/master/src/main.js#L56).

bennypowers: Would you like me to add a @since tag to the `withCredentials` docblock in main.js?

bennypowers:
@chrisn please take a look at the changes.

chrisn:
All looks good. This is a great contribution, thank you!

PR #219, opened 2018-02-23
Added pointer.mouseover and pointer.mouseleave events

markjongkind:
This PR adds support for two new Point events: `points.mouseover` and `points.mouseleave`.


markjongkind:
Great, you're welcome :)

chrisn:
I've had a closer look at this, and we should think about how the existing (undocumented) `pointDblClickHandler` and `pointDragEndHandler` callbacks should work. I like your approach of using events rather than callbacks, so maybe double-click and drag-end should be exposed as events? I wonder if we need drag-end at all.

Also, do we want to add equivalent events for segment start and end markers?

Issue #208, opened 2018-01-15
Memory leak. When is the "ready" event triggered in a peaks instance?


mwilde345:
I have code 
`    function addSegments() {
        //TODO: is this causing all the memory leaks....
        peaksInstance.on('peaks.ready', function () {
            peaksInstance.segments.add({
              stuff
            });
          });
          renderSegments(peaksInstance);
        });
      }`

that dynamically adds segments. When I traverse the waveform, moving the zoom bar, this code is continually run. thus I quickly flood the memory with create segments requests. This is giving me leak errors in the browser. 

So, how often is 'ready' triggered so I can figure out how to write this differently?

chrisn:
`peaks.ready` is emitted only once, after calling `Peaks.init()`, when the waveform data has been loaded and the views initialised. Where are you calling your `addSegments()` function from?

mwilde345:
The memory leak is saying there are too many event listeners. These event listeners are added every time I click inside the 'body' element as [written in your example page](https://github.com/bbc/peaks.js/blob/a1a22f9ffd0374fe808ff45c763e625f83160503/index.html#L271)

My mistake for not noticing that. I'm using it in a single page application where the audio is sent from the server and the peaks is initialized afterward on the same page. I will need to figure out how to add click functions for the items populated by [renderSegments](https://github.com/bbc/peaks.js/blob/a1a22f9ffd0374fe808ff45c763e625f83160503/index.html#L180)

mwilde345:
@chrisn Sorry I have to reopen because this is not behaving as if peaks is sending the 'ready' event just once.
I have a peaksInit function on my page:
`var peaksInstance=null;
function peaksInit(peaks){
peaksInstance = peaks.init(options);
peaksInstance.on('peaks.ready', function(){do stuff});
}`

well the 'do stuff' is triggered every time I add a segment or play a segment or drag the tracker over the waveform.
I removed all the body event listeners. This is stemming from the waveform changes.

should I not have a global peaksInstance variable?

chrisn:
If you can send me a complete and minimal code example that demonstrates the problem, I'll be happy to investigate.

mwilde345:
Seems like others have had my same issue. It doesn't stem from peaks, but Web Audio. I found this after testing with a .dat file

https://github.com/katspaugh/wavesurfer.js/issues/326
[...]

PR #230, opened 2018-06-29
When use `AudioContext`, only call requestAudioAndBuildWaveformData once

cky917:
When use `AudioContext` to generate the Peak,  and the `mediaSourceUrl` is null at the first time,  the `canplay` event listener will be bind. 
As we know, `canplay` event will be emit not only once, it can be emit so many times when we seek the video. 
https://github.com/bbc/peaks.js/blob/a7f11efe8fee6b40d9fadfb4ed17c47c68efde57/src/main/waveform/waveform.core.js#L179-L196
This code will cause `WaveformData` rebuild many times, we will not only lost the points marked before, but also each rebuild will add  many event listeners, may be memory leak at last. 
Add the `canplay` event listener by `once` can fix the problem maybe.

#208

### DIFF
--- a/src/main/waveform/waveform.core.js
+++ b/src/main/waveform/waveform.core.js
@@ -186,7 +186,7 @@
       );
     }
     else {
-      self.peaks.on('player_canplay', function(player) {
+      self.peaks.once('player_canplay', function(player) {
         self._requestAudioAndBuildWaveformData(
           player.getCurrentSource(),
           options.audioContext,


chrisn:
Thank you for this fix, great work! :-)

deniss-muhla:
Hello. Have same issue. Does this fix released to npm?

deniss-muhla:
It works like a charm! Thx!

chrisn:
@deniss-muhla Thanks for confirming. This is fixed in release 0.9.12. All credit goes to Keyi Chen.

PR #249, opened 2018-12-16
Feature: enable touch events for waveform container and pointer and segment markers on mobile devices

rfrei:
note:
- mouse events on mobile devices with connected USB mouse will only work with these Konva patches: https://github.com/rfrei/konva/commit/5bbb42d8fbca9c9d2e4e25c3b2f2323a6ca4cd10

### DIFF
--- a/src/main/views/helpers/mousedraghandler.js
+++ b/src/main/views/helpers/mousedraghandler.js
@@ -96,7 +96,7 @@ define(function() {
 
   MouseDragHandler.prototype.mouseMove = function(event) {
     var clientX = null;
-    
+
     if (event.type === 'touchmove') {
       clientX = parseInt(event.changedTouches[0].clientX, 10);
     }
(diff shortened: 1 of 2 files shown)


src/main/views/helpers/mousedraghandler.js
```diff
@@ -64,7 +65,12 @@ define(function() {
       return;
     }
 
-    this._mouseDownClientX = event.evt.clientX;
+    if (event.type === 'touchstart') {
+      this._mouseDownClientX = parseInt(event.evt.touches[0].clientX, 10);
```
chrisn: As `clientX` is already a number, I suggest using `Math.floor()` rather than `parseInt()`.

rfrei: Ok, thank you. I changed it.

chrisn:
Many thanks for contributing!

rfrei:
Thank you to for this nice software!
I have discovered another minor problem and will mention it in another issue: konvas _touchstart event handler does not call preventDefault() and because of that we get a second mousedown event after the touchstart event. This leads to short repeats of audio while playing and touching on the waveform. I come back soon.

Issue #222, opened 2018-05-12
Is it possible to hide the overview window?


bfeist:
I'd like to cause a zoomed window to open and play without the overview being visible. Is this possible via configuration, or will it be a customization? Thanks!

chrisn:
This is a planned feature, yes. See also #2, which requests disabling the zoomable view. I'm going to close this a duplicate, as the solution will allow you to choose which kind of view you want.

dpigera:
@bfeist Not sure if this helps, but removing it on render from the DOM also works:
```
this.$('.overview-container').remove();
```

chrisn:
Resolved in release 0.10.0. Please see https://github.com/bbc/peaks.js/blob/master/demo/zoomable-waveform.html for an example.

Issue #193, opened 2017-06-17
Simple music player - possible to remove advanced features - zooming and segmenting?


jurgenwerk:
I'm building a simple web music player app and since I'm generating waveform data using `audio_waveform-ruby` I thought about using `peaks.js` to display the waveform along with a simple play/pause interface and track progress indicator.

Basically, I'd like to replicate SoundCloud's player:

![image](https://user-images.githubusercontent.com/273660/27253588-d22e6626-5377-11e7-9304-1130f529e31a.png)

Is it possible to use a stripped-down version of peaks.js to implement a simple music player like this? How would you go about it?

And thanks for making peaks.js 🙏 .

chrisn:
Thanks for the kind feedback!

Peaks.js is currently designed to have the two waveform views you see in this [screenshot image](https://raw.githubusercontent.com/bbc/peaks.js/master/peaks.png). There are other projects out there that are closer to what you want, are there particular reasons you prefer Peaks.js?

That said, there is [work underway](https://github.com/bbc/peaks.js/pull/177) to restructure the code to support a single view, but it's slow progress due to the time I have available.

If you want to do it yourself, you could always take the Peaks.js code and strip out all the functionality you don't need. There's code in [this pull request](https://github.com/bbc/peaks.js/pull/181) to implement the highlight (orange) region.

chrisn:
Release 0.10.0 allows you to create a single waveform view. Please see https://github.com/bbc/peaks.js/blob/master/demo/overview-waveform.html folder for an example.

PR #258, opened 2019-03-06
Added 'containers' option

chrisn:
This pull request adds a `containers` option that allows users to select whether the overview waveform, zoomable waveform, or both, are rendered. See #2, #222, #53, #193, #46.


zachsa:
@chrisn - this looks pretty helpful! I would also like to be able to toggle zoomed/normal container programmatically.

So the use case I'm thinking of would be a 'show more' button that then displayed  the zoomable section.

Is it possible to choose which of the containers appears on top? (at the moment the zoomed container is always on top?)

zachsa:
> @chrisn - this looks pretty helpful! I would also like to be able to toggle zoomed/normal container programmatically.
> 
> So the use case I'm thinking of would be a 'show more' button that then displayed the zoomable section.
> 
> Is it possible to choose which of the containers appears on top? (at the moment the zoomed container is always on top?)

But also... if these were useful, I think that should be a separate pull request. Being able to hide one or another container would be immediately useful.

chrisn:
> I would also like to be able to toggle zoomed/normal container programmatically.

Yes, let's follow this up in a separate issue / PR.

> Is it possible to choose which of the containers appears on top? (at the moment the zoomed container is always on top?)

This is controlled by your page layout and styling. If you want the zoomed container to appear below, you can order the `div`s as needed, or apply styling to position them:

```html
<div id="waveform-container">
  <div id="overview-container"></div>
  <div id="zoomview-container"></div>
</div>
```

Issue #247, opened 2018-11-23
Is it possible to update a segment's labelText property programmatically?


zachsa:
I'm using React, and I initiate a segment with `labelText` as a reference to state within a component:

```
labelText: this.state.text
```

But updating the component state doesn't trigger the label of the segment to update. And if i retrieve the segment and update the labelText field then the changed label text doesn't reflect on the waveform.

Is it possible to force the labelText to redraw without re-rendering the waveform?

zachsa:
I would be interested in working on this. Please let me know what I should look into and how to proceed. Thanks!

chrisn:
Hi Zach,

Thank you! Here are a few notes, I hope these are helpful...

Segments are represented by [`Segment`](https://github.com/bbc/peaks.js/blob/master/src/main/markers/segment.js) objects, and these are managed by a [WaveformSegments](https://github.com/bbc/peaks.js/blob/master/src/main/markers/waveform.segments.js) object, which provides the `peaks.segments` API methods.

We want to be able to modify any of the segment properties and for those changes to be reflected in the UI.

The first step would be to modify `Segment` to store a reference to the `WaveformSegments` object, so we can notify it when things change:

```javascript
function Segment(parent, id, startTime, endTime, labelText, color, editable) {
  this._parent = parent; // leading underscore for private variables
  // etc
}
```

Then we can add methods to allow value changes. Rather than adding `getStartTime()` and `setStartTime()` methods we could use JavaScript [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) and [setter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) syntax. This would also allow us to prevent changes to the `id` field, which must not be changed after the `Segment` is created.

The `WaveformSegments` object emits events to cause the UI to update. The UI is managed by the [`SegmentsLayer`](https://github.com/bbc/peaks.js/blob/master/src/main/views/segments-layer.js) object, which listens for those events.

We currently have `segments.add`, `segments.remove`, and `segments.remove_all` events. I suggest adding a `segments.update` event for segment value changes. The event handlers are set up in [`SegmentsLayer._registerEventHandlers()`](https://github.com/bbc/peaks.js/blob/master/src/main/views/segments-layer.js#L48).

[...]

PR #260, opened 2019-06-08
Pull request issue 247

zachsa:
1. Updated Segment API to allow for programmatic segment property updates
2. Updated test suite to cover validated, programmatic segment updates

### DIFF
--- a/peaks.js.d.ts
+++ b/peaks.js.d.ts
@@ -11,6 +11,7 @@ declare module 'peaks.js' {
     color?: string;
     labelText?: string;
     id?: string;
+    update: (options: Object) => void;
   }
 
   interface Point {
@@ -19,6 +20,7 @@ declare module 'peaks.js' {
     color?: string;
     labelText?: string;
     id?: string;
+    update: (options: Object) => void;
   }
 
   interface RequiredOptions {
--- a/peaks.js.d.ts
+++ b/peaks.js.d.ts
@@ -4,6 +4,15 @@
  */
 
 declare module 'peaks.js' {
+
+  interface SegmentUpdateOptions {
+    startTime?: number;
+    endTime?: number;
+    editable?: boolean;
+    color?: string;
+    labelText?: string;
+  }
+
   interface Segment {
     startTime: number;
     endTime: number;
@@ -11,7 +20,14 @@ declare module 'peaks.js' {
     color?: string;
     labelText?: string;
     id?: string;
-    update: (options: Object) => void;
+    update: (options: SegmentUpdateOptions) => void;
+  }
+
+  interface PointUpdateOptions {
+    time?: number;
+    labelText?: string;
+    color?: string;
+    editable?: boolean;
   }
 
   interface Point {
@@ -20,7 +36,7 @@ declare module 'peaks.js' {
     color?: string;
     labelText?: string;
     id?: string;
-    update: (options: Object) => void;
+    update: (options: PointUpdateOptions) => void;
   }
 
   interface RequiredOptions {
--- a/src/main/markers/waveform.segments.js
+++ b/src/main/markers/waveform.segments.js
@@ -126,32 +126,8 @@ define([
       throw new TypeError('peaks.segments.add(): expected a Segment object parameter');
     }
 
-    if (!Utils.isValidTime(options.startTime)) {
-      // eslint-disable-next-line max-len
-      throw new TypeError('peaks.segments.add(): startTime should be a valid number');
-    }
-
-    if (!Utils.isValidTime(options.endTime)) {
-      // eslint-disable-next-line max-len
-      throw new TypeError('peaks.segments.add(): endTime should be a valid number');
-    }
-
-    if (options.startTime < 0) {
-      // eslint-disable-next-line max-len
-      throw new RangeError('peaks.segments.add(): startTime should not be negative');
-    }
-
-    if (options.endTime < 0) {
-      // eslint-disable-next-line max-len
-      throw new RangeError('peaks.segments.add(): endTime should not be negative');
-    }
-
-    if (options.endTime <= options.startTime) {
-      // eslint-disable-next-line max-len
-      throw new RangeError('peaks.segments.add(): endTime should be greater than startTime');
-    }
-
     var segment = new Segment(
+      this,
       Utils.isNullOrUndefined(options.id) ? this._getNextSegmentId() : options.id,
       options.startTime,
       options.endTime,
--- a/src/main/views/segments-layer.js
+++ b/src/main/views/segments-layer.js
@@ -49,10 +49,10 @@ define([
     var self = this;
 
     this._peaks.on('segments.update', function(segment) {
+      var redraw = false;
       var segmentGroup = self._segmentGroups[segment.id];
       var frameOffset = self._view.getFrameOffset();
       var width = self._view.getWidth();
-      var redraw = false;
       var frameStartTime = self._view.pixelsToTime(frameOffset);
       var frameEndTime   = self._view.pixelsToTime(frameOffset + width);
 
--- a/src/main/views/segments-layer.js
+++ b/src/main/views/segments-layer.js
@@ -51,7 +51,9 @@ define([
     this._peaks.on('segments.update', function(segment) {
       var segmentGroup = self._segmentGroups[segment.id];
 
-      segmentGroup.label.setText(segment.labelText);
+      if (segmentGroup) {
+        segmentGroup.label.setText(segment.labelText);
+      }
     });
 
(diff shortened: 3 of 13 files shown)


zachsa:
This pull request addresses issue #247

src/main/markers/waveform.segments.js
```diff
@@ -126,32 +126,8 @@ define([
...
-    }
-
-    if (options.endTime <= options.startTime) {
-      // eslint-disable-next-line max-len
-      throw new RangeError('peaks.segments.add(): endTime should be greater than startTime');
-    }
-
     var segment = new Segment(
+      this,
       Utils.isNullOrUndefined(options.id) ? this._getNextSegmentId() : options.id,
```
zachsa: This could also be moved into segment.js since segments now reference parent objects. Then the waveform-segments object doesn't need to reference Utils at all.

src/main/views/segments-layer.js
```diff
@@ -48,6 +48,14 @@ define([
   SegmentsLayer.prototype._registerEventHandlers = function() {
     var self = this;
 
+    this._peaks.on('segments.update', function(segment) {
```
chrisn: We'll need to also update the UI objects if the `startTime` or `endTime` changes. This is a bit more complex than changing the label, as a change to the start/end times could affect whether a segment remains (or becomes) visible in the waveform view.

test/unit/segment-spec.js
```diff
@@ -1,59 +1,141 @@
...
+      var newLabelText = 'new label text';
+      var segment = p.segments.getSegments()[0];
+
+      segment.labelText = newLabelText;
+      expect(segment.labelText).to.equal(newLabelText);
+    });
+  });
+
+  describe('startTime', function() {
+    it('should be possible to update programatically', function() {
```
chrisn: I just realised an issue with changing the start and end times. If you have a segment, for example at (5, 10) and you want to move it to (15, 20), you have to update the end time first, otherwise you get an exception. I wonder if we should prefer to have an `updateTime(start, end)` method to be able to set both together? Or maybe to flip the start and end times automatically if you set the end time after the start time?

zachsa: ah okay. I think it would make sense if the startTime/endTime were always updated together in this case. So NOT providing a means of updating starTime and endTime, and instead providing a setter function `updateTime(start, end)`?

zachsa: This should be resolved now

src/main/views/segments-layer.js
zachsa: Looking at the event handler for the `segments.add` event, I see this code:

```
this._peaks.on('segments.add', function(segments) {
  var frameOffset = self._view.getFrameOffset();
  var width = self._view.getWidth();

  var frameStartTime = self._view.pixelsToTime(frameOffset);
  var frameEndTime   = self._view.pixelsToTime(frameOffset + width);

  segments.forEach(function(segment) {
    if (segment.isVisible(frameStartTime, frameEndTime)) {
      self._addSegmentGroup(segment);
    }
  });

  self.updateSegments(frameStartTime, frameEndTime);
});
```

It looks like I should just include the same logic in the `segments.updated` event handler?

zachsa: i.e... the value of `isVisible` needs to be reset if necessary, and then `self.updateSegments` should be called with new frameStartTime, frameEndTime?

test/unit/segment-spec.js
chrisn: I wonder if the API would be simpler if we added a `segment.update()` method, that takes an object with the values you want to change, rather than use getter and setters. For example:

```javascript
segment.update({ labelText: 'New label' });
segment.update({ labelText, 'New label', startTime: 20, endTime: 30 });
```

What do you think?

src/main/views/segments-layer.js
chrisn: Because `segments.updated` will only receive one segment, we only need to check that particular segment, rather than all of them, so I would avoid calling `updateSegments`. I guess a way to implement would be to remove the segment if it's currently visible, then add it in its new position, maybe something like this:

```javascript
var redraw = false;

var segmentGroup = this._segmentGroups[segment.id];

if (segmentGroup) {
  this._removeSegment(segment);
  redraw = true;
}

var frameOffset = self._view.getFrameOffset();
var width = self._view.getWidth();

var frameStartTime = self._view.pixelsToTime(frameOffset);
var frameEndTime   = self._view.pixelsToTime(frameOffset + width);

if (segment.isVisible(frameStartTime, frameEndTime)) {
  this._addSegmentGroup(segment);
  redraw = true;
}

if (redraw) {
  this._layer.draw();
}
```

test/unit/segment-spec.js
zachsa: I agree that is a better approach

zachsa: This is added in commit #6f32044

zachsa:
@chrisn - I think I have addressed everything you raised. Let me know if you have any more comments. I also added a `Segment API` section to the documentation. You can view this in my forked repository here: https://github.com/zachsa/peaks.js

test/unit/segment-spec.js
chrisn: Ah yes, we don't currently use any ES6 features... although I would like to modernise the code at some point.

chrisn:
Many thanks, this looks really good. I'll do some tests. Would you mind including the documentation changes in the pull request?

The only other thing I would like to do is modify the Points API in the same way, for consistency. I'm happy to do that, if you prefer?

zachsa:
I'll have a look at the Point API - should be pretty quick. The documentation changes are already in the pull request.

chrisn:
Thanks! One final task is to update the TypeScript declarations file with the API changes: https://github.com/bbc/peaks.js/blob/master/peaks.js.d.ts

zachsa:
> Many thanks, this looks really good. I'll do some tests. Would you mind including the documentation changes in the pull request?
> 
> The only other thing I would like to do is modify the Points API in the same way, for consistency. I'm happy to do that, if you prefer?

I'm struggling to get the points to redraw correctly using the same approach as is used for the segments. This doesn't work:

```
    this._peaks.on('points.update', function(point) {
      var redraw = false;
      var pointGroup = self._pointGroups[point.id];
      var frameOffset = self._view.getFrameOffset();
      var width = self._view.getWidth();
      var frameStartTime = self._view.pixelsToTime(frameOffset);
      var frameEndTime   = self._view.pixelsToTime(frameOffset + width);

      if (pointGroup) {
        self._removePoint(point);
        redraw = true;
      }

      if (point.isVisible(frameStartTime, frameEndTime)) {
        self._addPointGroup(point);
        redraw = true;
      }

      if (redraw) {
        self._layer.draw();
      }
    });
```

With this code, I have to 'jiggle' the waveform to get the point to display.

this code works:

```
    this._peaks.on('points.update', function(point) {
      var frameOffset = self._view.getFrameOffset();
      var width = self._view.getWidth();
      var frameStartTime = self._view.pixelsToTime(frameOffset);
      var frameEndTime   = self._view.pixelsToTime(frameOffset + width);

      self._removePoint(point);
      if (point.isVisible(frameStartTime, frameEndTime)) {
        self._addPointGroup(point);
      }

      self.updatePoints(frameStartTime, frameEndTime);
    });
```

Is it okay?

zachsa:
Point API is adjusted to match the Segment API. Also the updated the test suite and the documentation, similarly to the changes for the segment API. The readme file is viewable on my forked repository in master branch (and is included in the pull-request for this repo).

(https://github.com/zachsa/peaks.js#point-api)

zachsa:
> Thanks! One final task is to update the TypeScript declarations file with the API changes: https://github.com/bbc/peaks.js/blob/master/peaks.js.d.ts

I think I have done this... Please let me know if I haven't done this correctly as I haven't really worked with typescript or typescript definitions before.

peaks.js.d.ts
```diff
@@ -11,6 +11,7 @@ declare module 'peaks.js' {
     color?: string;
     labelText?: string;
     id?: string;
+    update: (options: Object) => void;
```
chrisn: Using object is proabably too generic. I'd suggest:
```typescript
update: (options: SegmentUpdateValues) => void;
```
and
```typescript
interface SegmentUpdateValues {
  startTime?: number;
  endTime?: number;
  editable?: boolean;
  color?: string;
  labelText?: string;
}
```
(and similar for `Point.update`)

zachsa: done

chrisn:
All good, so I have merged to `master`. Thank you for all your work on this!

zachsa:
Thank you!

PR #267, opened 2019-06-24
allow glob for testing

### LINKED ISSUE #266: question: run a single test?
out of curiosity, how do you run a single test? either I'm missing something or it seems to be all or nothing..

gmarinov:
close #266 - allows running selected test suites by command line argument

examples:

```
npm test -- --glob=api-player-spec.js
```

```
npm run test-watch -- --glob=api-player-spec.js
```

```
npm test -- --glob=waveform*.js
```

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,4 +1,5 @@
 'use strict';
+/* eslint-disable */
 
 function filterBrowsers(browsers, re) {
   return Object.keys(browsers).filter(function(key) {
@@ -9,6 +10,7 @@ function filterBrowsers(browsers, re) {
 module.exports = function(config) {
   var isCI = Boolean(process.env.CI) && Boolean(process.env.BROWSER_STACK_ACCESS_KEY);
 
+  var glob = config.glob || '**/*.js'
   // Karma configuration
   config.set({
     // The root path location that will be used to resolve all relative paths
@@ -38,7 +40,7 @@ module.exports = function(config) {
       { pattern: 'test/test_img/*', included: false },
       { pattern: 'test_data/*', included: false },
       { pattern: 'test_data/sample.{dat,json}', included: false, served: true },
-      { pattern: 'test/unit/**/*.js', included: true }
+      { pattern: 'test/unit/'+glob, included: true }
     ],
 
     mime: {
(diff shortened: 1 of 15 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ docs
 node_modules
 peaks.js
 peaks.js.map
+.idea/
```
chrisn: This seems to be specific to your development environment. Please use your [global .gitignore](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore) instead.

karma.conf.js
```diff
@@ -1,4 +1,5 @@
 'use strict';
+/* eslint-disable */
```
chrisn: We should prefer to fix any lint errors, rather than disabling linting.

test/unit/api-player-spec.js
```diff
@@ -33,7 +34,8 @@ describe('Peaks.player', function() {
 
     it('should return an updated time if it has been modified through the audio element', function(done) {
       p.on('player_seek', function(currentTime) {
-        var diff = Math.abs(p.player.getCurrentTime() - newTime);
+        // var diff = Math.abs(p.player.getCurrentTime() - newTime);
```
chrisn: This test is intended to check `player.getCurrentTime()`. We should add separate test to also check `currentTime` as reported in the `player_seek` event.

gmarinov: ugh, sorry, I didn't notice this, it was just me testing..

.gitignore
gmarinov: TIL, thanks

karma.conf.js
gmarinov: I assumed this, being outside `src/` would be exempt. fixed

chrisn: Ah, I think the eslint command in package.json isn't picking it up. I'll fix this.

gmarinov: in f8b9712 I modified it slightly so the linter stopped complaining (at least on mine)

test/unit/api-player-spec.js
gmarinov: (p.s. reverted)

PR #268, opened 2019-06-27
cue events

### LINKED ISSUE #243: Built in event for when playhead hits a point/segment ?
I'd like to be able to play a sound when the play head hits a point and/or segment.
Is it possible to subscribe to a built in event for this or will it require a custom one?

thanks

gmarinov:
close #243 - add CueEmitter to track events from playhead crossing point or segment boundaries.

- emits only while playing, does not emit while seeking.
- listens for changes to points and segments and updates accordingly
- detaches and releases its internal reference if Peaks instance `.destroy()` is called

maybe of interest to @peet-j

`demo/cue-events.html`

cue-events-redo

#### API:
```
const peaks = Peaks.init({...,  emitCueEvents: true });
peaks.on('points.enter', console.log)
peaks.on('segments.enter', console.log)
peaks.on('segments.exit', console.log)
```

#### notes:
- checked on Chrome 75 macOS, Firefox 67 macOS, Safari iOS 12, Safari 12.1 macOS
- has some room for optimisation in `CueEmitter.prototype._onUpdate` should a need arise
- mostly external to the main Peaks class, keeps a bit of extra state to avoid changes to Peaks' internals

#### adds tests:

```
  CueEmitter
    ✓ should initialise correctly
    ✓ should initialise with already existing points
    ✓ should destroy tracker when peaks is destroyed
    points -> marks
      ✓ should add marks when point is added
      ✓ should reorder when point is added earlier
      ✓ should update marks when point is updated
      ✓ should remove marks when point is removed
      ✓ should update when all points are removed
    segments -> marks
      ✓ should add dual marks when segment is added
      ✓ should reorder when segment is added earlier
      ✓ should update marks when segment is updated
      ✓ should remove marks when segment is removed
      ✓ should update when all segments are removed
    events
      ✓ should update internal previous time when seeking
      ✓ should emit point events in forward
      ✓ should emit point events in reverse
      ✓ should emit segment events in forward
      ✓ should emit segment events in reverse

  CueMark
    ✓ should not allow overwriting props
    ✓ should emit correct events for POINT
    ✓ should emit correct events for SEGMENT_START
    ✓ should emit correct events for SEGMENT_END
```


chrisn:
Thanks for this contribution. This is going to take a while to review, so please bear with me.

chrisn:
My initial thought: I wonder if it's possible to implement this using the HTML `TextTrack` API, if we create `VTTCue` objects for each point and segment.

gmarinov:
I wasn't sure about how far browser support should extend, and feeling uneasy about having CueEmitter manage a more complex state that isn't the source of truth anyway. Sounds like a neat idea for the source state? Warrants discussion to size it up at least, could be a massive change to Peaks' internals?

chrisn:
I just pushed an implementation using TextTrack and VTTCue: https://github.com/bbc/peaks.js/compare/text-track-cue-events.

This approach works well, but has two main issues:

* The timing of enter/exit event firings varies between browsers, due to the rate at which [time marches on](https://html.spec.whatwg.org/multipage/media.html#time-marches-on) is specified to run.
* If I create a segment (from start time _ts_ to end time _te_), then place the playhead marker at time _tp_, where _ts_ < _tp_ < _te_, and then adjust the segment end time such that _te_ < _tp_, the segment's exit event is not fired. (This also happens with the implementation in this PR.)

gmarinov:
Enter/exit events not firing consistently when an overriding user action messes with the segment definition or re-seeks the transport is still within reason, no? Likely listener's expectations.

With regards to timing's resolution, I do find it too coarse and the timing is often off, that's why I wired the emitter in this branch to run on animation frame and only fall back on `player_time_update` when animation framerate drops (i.e. window isn't visible).

Minor difference in events in cue/vtt branch, points fire enter and exit event pairs.
Still a choice between two flavours of additional state 😄 

Up to you..

README.md
```diff
@@ -711,6 +711,17 @@ const view = instance.views.getView('zoomview');
 view.setAmplitudeScale(1.0);
 ```
 
+## Cue events
+
+Emit events when the playhead reaches a point or segment boundary:
+
+```js
+const cues = new Peaks.CueEmitter(instance);
```
chrisn: I would prefer to extend the existing API, rather than introduce a new class

```javascript
peaks.on('segments.enter', function(segment) { ... });
peaks.on('segments.exit', function(segment) { ... });
peaks.on('points.enter', function(point) { ... });
```

README.md
```diff
@@ -711,6 +711,17 @@ const view = instance.views.getView('zoomview');
...
+cues.on('cue.point', function(id, time) { /* ... */ });
```
chrisn: I would prefer to use `points.enter`, `segments.enter`, `segments.exit` as event names, being more similar to the  TextTrackCue API.

Also, I think the handler functions should receive a `Point` or `Segment` object as parameter, rather than `id` and `time`.

src/main/cues/cue-emitter.js
```diff
@@ -0,0 +1,224 @@
...
+      end = marks.length;
+      step = 1;
+    }
+    else {
+      start = marks.length - 1;
+      end = -1;
+      step = -1;
+    }
+    // marks are sorted
+    for (var i = start, mark, markTime; isForward ? i < end : i > end; i += step) {
```
chrisn: As you mentioned, this is expensive when there's a large number of points and segments.

src/main/cues/cue-emitter.js
```diff
@@ -0,0 +1,224 @@
...
+  // when the window isn't in focus, rAF is throttled
+  // falling back to timeUpdate
+
+  CueEmitter.prototype.onTimeUpdate = function(time) {
+    if (windowIsVisible()) {
+      return;
+    }
+    var el = getMediaElement(this.peaks);
+
+    if (this.peaks.player.isPlaying() && !el.seeking) {
```
chrisn: We can add a `peaks.player.isSeeking()` method.

src/main/cues/cue-emitter.js
```diff
@@ -0,0 +1,224 @@
...
+
+  CueEmitter.prototype.attach = function() {
+    var peaks = this.peaks;
+
+    peaks.on('player_time_update', this.onTimeUpdate);
+    peaks.on('player_play', this.onPlay);
+    peaks.on('player_seek', this.onSeek);
+    peaks.on('destroyed', this.onPeaksDestroyed);
+    for (var i = 0; i < triggerUpdateOn.length; i++) {
+      peaks.on(triggerUpdateOn[i], this._updateMarks);
```
chrisn: These operations get more expensive with increasing numbers of points and segments.

gmarinov: redoing the marks is triggered by a user action, it's outside the nominal (hot) loop when playing? I wanted to bind it to `points.update` and `segments.update` only but it seems the UI modifies them directly.

src/main/cues/cue-emitter.js
gmarinov: Theoretically, yes. I guess it could be optimised further by having a smarter data structure (ideas welcome), but I feel the feature/UI itself would have become unusable from having too many points or segments, before the path actually runs slow.

gmarinov:
addressed comments:
- renamed events 
- added `player.isSeeking()` method
- updated docs
also refactored the cue marks to be objects now, to keep things simple.

README.md
```diff
@@ -798,6 +801,17 @@ const view = instance.views.getView('zoomview');
 view.showPlayeadTime(false); // Remove the time from the playhead marker.
 ```
 
+## Cue events
+
+Emit events when the playhead reaches a point or segment boundary:
+
+```js
+const peaks = Peaks.init({ ..., emitCueEvents: true });
+peaks.on('points.enter', function(point, time, id) { /* ... */ });
```
chrisn: The `time` and `id` parameters aren't needed here, as you can get at them via the `point` or `segment` object.

README.md
```diff
@@ -272,6 +272,9 @@ var options = {
 
   // async logging function
   logger: console.error.bind(console),
+  
+  // if true, emit cue events on the peaks instance (see Cue Events) 
+  emitCueEvents: false,
```
chrisn: :+1:

README.md
```diff
@@ -850,6 +864,24 @@ Peaks instances emit events to enable you to extend its behaviour according to y
...
+When the playhead reaches a point or segment boundary, a cue event is emitted.
+   
+| Event name               | Arguments                |
+| ------------------------ | ------------------------ |
+| `points.enter`           | `Point point, Number time, String id` |
+| `segments.enter`         | `Segment segment, Number time, String id` |
+| `segments.exit`          | `Segment segment, Number time, String id` |
+
+When the media element's `.playbackRate` is set to negative (playing backwards on supported browsers only), 
+segment events are emitted in reverse (i.e. `.in` on the `segmentEnd` boundary and `.out` on the `segmentStart` boundary).
```
chrisn: Minor nit: the event names are enter/exit

demo/cue-events.html
```diff
@@ -0,0 +1,332 @@
+<!DOCTYPE html>
```
chrisn: We'll need to update this page, as I fixed a bug with the point/segment input fields in https://github.com/bbc/peaks.js/commit/5448e420bfb4e053a58f67136673af5afe67eb7b.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "peaks.js",
-  "version": "0.10.1",
+  "version": "0.11.0",
```
chrisn: Please don't bump the version number, as merging this commit would automatically trigger publication of the new version to npm. I'll do that when I prepare the next release, which will include some other changes.

test/unit/cue-emitter-spec.js
```diff
@@ -0,0 +1,251 @@
...
+
+  afterEach(function() {
+    if (p) {
+      p.destroy();
+    }
+    p = undefined;
+    cue = undefined;
+  });
+
+  it('should initialise correctly', function(done) {
```
chrisn: We can remove the `done` parameter, as these tests aren't async.

src/main.js
```diff
@@ -490,7 +500,10 @@ define('peaks', [
     if (this.player) {
       this.player.destroy();
     }
+    this.emit('destroyed');
   };
 
+  Peaks.CueEmitter = CueEmitter;
```
chrisn: I'd prefer to avoid exporting `CueEmitter`, it's an implementation detail that shouldn't be exposed to users.

src/main/cues/cue-emitter.js
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * @class
+   * @property {Peaks} peaks
+   * @property {Array.<CueMark>} marks
+   * @param {Peaks} peaks
+   * @constructor
+   */
+  function CueEmitter(peaks) {
+    this.marks = Array();
```
chrisn: We're moving to a naming convention where private fields have a leading underscore. If you prefer, I can make this change?

src/main/cues/cue-emitter.js
chrisn: Some of our use cases involve having lots of segments (e.g., annotating speech vs music segments), although in this case we may not be using the events. We'd need to do some measurements to see if optimisation is needed - but if so, I'd expect a binary search approach would be faster, for larger numbers of events.

src/main.js
gmarinov: ah yes, my bad, forgot it there..

src/main/cues/cue-emitter.js
gmarinov: mm, yep, i thought not to optimise it prematurely. I think any slowdown here would be more because of a GC issue, and the worst i can think of is when dragging events are emitted at a very high rate with thousands of segments defined (in which case other parts of the current UI might struggle just the same)

package.json
gmarinov: I took it out

src/main/cues/cuemark.js
```diff
@@ -0,0 +1,88 @@
...
+   * @property {number} time
+   * @property {string} id
+   * @property {number} type
+   * @constructor
+   * @param {number} time
+   * @param {number} typ
+   * @param {string} id
+   */
+  function CueMark(time, typ, id) {
+    Object.defineProperties(this, {
```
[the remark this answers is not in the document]
gmarinov: → adffb14

src/main/cues/cue-emitter.js
chrisn: > I wanted to bind it to points.update and segments.update only but it seems the UI modifies them directly.

This looks like an oversight, editing points/segments via the UI should probably emit `update` events.

chrisn:
I have made a few changes on your branch. I hope this is OK.

I added a (currently failing) test for events that should be emitted on seeking. To fix this, we'd probably need to keep track of which segments are active, then update and compare this after a seek to emit the right enter/exit events.

test/unit/cue-emitter-spec.js
```diff
@@ -262,5 +262,57 @@ describe('CueEmitter', function() {
 
       cue._onUpdate(1.1, 1.0);
     });
+
+    it('should emit events on seeking', function(done) {
```
gmarinov: if it should emit on seeking really is the question for me. i deliberately made it so it doesn't - https://github.com/bbc/peaks.js/blob/9973a32d0cc91aa36d6c6f34cd36aed8857e90a3/src/main/cues/cue-emitter.js#L142 - likely the only change it needs to actually start emitting this stuff again (though the tests might need some more mocking of the transport)

my position is: it shouldn't emit on seeking. my interpretation of seeking is that the playhead didn't touch the material, and these events come from the playhead. the listener (not the emitter) must adjust its state about what region it's in, as soon as seeking has occurred. this is trivial for the listener which has access to the list of regions and new playhead position/time.

chrisn: Sorry for the delay on this. I have a working implementation that emits events on seek,

chrisn:
Thank you @gmarinov :100: :+1:

PR #276, opened 2019-09-13
Updated README to include install note

rramphal:
This is a minor README change to let the user know that they can't run the demos with (`npm start`) without installing dependencies first (`npm install`).


chrisn:
Good catch, thanks for fixing this.

PR #278, opened 2019-09-20
Segment dragstart dragend events

### LINKED ISSUE #277: Marker identification for segment drag events
When a user drags an editable segment in/out marker, there is no way to know which marker they are dragging. We only receive the updated segment information. This is problematic for some situations for example when wanting to see a video preview or thumbnail as a marker is dragged. If the in marker is being dragged, you want the video preview/thumbnail to be the start of the in/out range. If the end marker is being dragged, you want the end.

It would be great if the event included some sort of identifier that indicated whether the in marker or out marker is being dragged.

Spidy88:
Adds `dragstart` and `dragend` event to segments.
Adds inMarker and outMarker identification to `dragstart`, `dragend`, and `dragged` events

Resolves #277

### DIFF
--- a/src/main/views/segments-layer.js
+++ b/src/main/views/segments-layer.js
@@ -229,7 +229,7 @@ define([
    * @param {Segment} segment
    */
 
-  SegmentsLayer.prototype._onSegmentHandleDrag = function(segmentGroup, segment) {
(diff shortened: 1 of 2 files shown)


chrisn:
Thank you for this, and sorry for taking so long to respond. This is a good improvement. I'll merge this add documentation for the `segments.dragstart` and `segments.dragend` events.

Issue #224, opened 2018-05-17
rerender waveform when height/width of peaks.js container has been changed


ojczeo:
I would like to make `peaks.js` more resoponsive - rerender waveform when height/width of peaks container has been changed.

Is there any way to achieve that? for now as I can see there is only way to call 
`peaksInstance.waveform.onResize();` - but it only rerenders width.

chrisn:
It would be great to implement this. I'll take a look when I have time, but will happily accept a pull request!

ojczeo:
@chrisn ok, will try to explore and find solution, and PR as the result

chrisn:
Peaks.js v0.18.0 added a [`view.fitToContainer()`](https://github.com/bbc/peaks.js/blob/master/README.md#viewfittocontainer) method that rerenders the waveform to fit the container's width and height. Feedback is welcome!

PR #288, opened 2019-11-06
Overview highlight offset

jodonnell:


chrisn:
Thanks! Perhaps we should also add code to prevent the height from going negative?

jodonnell:
well we could do 
```y: Math.max(this._options.overviewHighlightOffset, 0),```
although going negative won't harm anything because it will draw outside the canvas which won't get rendered.

Or were you thinking like at startup time around where the deprecation warnings are it would throw an error?

chrisn:
I was more thinking of possibility that Konva may error, but if that doesn't happen we're fine. I don't think we need a warning.

jodonnell:
ahh dang, i didn't update the README.

should I make one more PR with the Readme updated?

And if i want to create a release after it suggests I push right to master but i really doubt i have permissions to do that, so I could create a PR were I update the change log and bump the version.

chrisn:
Don't worry, I'll do the README and the release. Thanks for contributing!

jodonnell:
Okay great, thanks for the help.

chrisn:
I have just pushed the v0.14.3 release, it should be available in npm shortly.

chrisn:
So we do need to add some validation, if I make the offset too large we get this error from Konva:

> DOMException: Failed to execute 'arc' on 'CanvasRenderingContext2D': The radius provided (-7.5) is negative.

jodonnell:
okay i'll take a look

jodonnell:
okay this happens when the offset is over half the height.

i'm not sure how to implement this though, because where i would think to put this validation, the same place i put the depreciation warning won't work i don't think.

the height is often set via css so i don't think i'd have access to the true height of the canvas.

where do you think putting this would make sense?

chrisn:
Another consideration is dynamically resizing of the waveform view (see #224). What I'd suggest is not throwing an error, but instead applying some minimum values to `y` and `height` in `_createHighlightRect`, which override the user's option setting.

jodonnell:
good thinking 👍 , will have pr up soon

jodonnell:
https://github.com/bbc/peaks.js/pull/289

PR #289, opened 2019-11-06
Bound the highlight rectangle to prevent an error

jodonnell:


src/main/views/waveform.overview.js
```diff
@@ -270,14 +270,17 @@ define([
 
     var startOffset = this.timeToPixels(startTime);
     var endOffset   = this.timeToPixels(endTime);
+    var offset = Math.max(this._options.overviewHighlightOffset, 0);
+
+    offset = Math.min(offset, Math.floor(this._height / 2));
```
chrisn: We have a [`Utils.clamp`](https://github.com/bbc/peaks.js/blob/master/src/main/waveform/waveform.utils.js#L99) function that we could use to simplify this.

jodonnell: updated

Issue #291, opened 2019-11-11
Missing Events "segments.dragstart" and "segments.dragend" in Typescript definition


tscz:
Hi @chrisn,

While trying to register a callback on the "segments.dragend" event in a Typescript setting I realized that it is not yet declared in the definition file and thus compilation fails for
```javascript
peaks.on("segments.dragend", (segment: Segment, inMarker : boolean) => {
   console.log("segments.dragend:", segment, inMarker);
});
```
with
`Argument of type '"segments.dragend"' is not assignable to parameter of type '"peaks.ready" | "points.add" | "points.dblclick" | "points.dragend" | "points.dragmove" | "points.dragstart" | "points.mouseenter" | "points.mouseleave" | "points.remove_all" | ... 13 more ... | "user_seek"'.`

Double-checked against the doc and code.

Missing events:
- segments.dragstart
- segments.dragend

Events with missing parameter:
- segments.dragged (boolean inMarker missing)

tscz:
Also shouldn't the return types of all callbacks be of type "void" instead of "any" as correct API contract in the [definition](https://github.com/bbc/peaks.js/blob/master/peaks.js.d.ts)?

See from the eventemitter2 definition:
https://github.com/EventEmitter2/EventEmitter2/blob/a7ea0eded13b2fa267d90317824db839bd11c331/eventemitter2.d.ts#L30

And the recommendation from the Typescript team:
https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#return-types-of-callbacks

chrisn:
I think you're right. Many thanks for contributing a fix for this.

PR #292, opened 2019-11-11
Pull request issue 291

tscz:
Updates Typescript definition:
1. add event "segments.dragstart"
2. add event "segments.dragend"
3. update event "segments.dragged"
4. change return type of callbacks to void

related to issue #291


chrisn:
Thank you!

PR #306, opened 2020-02-02
Update typescript definitions for 0.18.0 changes

tscz:
Changes from 0.18.0. were missing:

Views API:
* destroyOverview()
* destroyZoomview()

View API:
* setZoom
* setStartTime

Further Improvements:
* name parameter in views.getView(name) is checked at compile time
* options parameter in setZoom(options) is checked at compile time. (Either scale or seconds is given and is either a number or string "auto").


chrisn:
Thanks! :-)

PR #308, opened 2020-02-03
Allow using raw peak data to render waveform

is343:
Peaks.js is using [waveform-data.js](https://github.com/bbc/waveform-data.js) to render the waveform. Currently the only method to pass the pre-rendered peaks to `WaveformData.create` is to first fetch it from the server. 

I have added an option to supply the raw peaks data as `rawData` to use when creating the peaks. 
In my case, I am rendering extra waveforms on the page using waveform-data.js. But, this requires me to fetch the peaks twice, once via peaks.js and once to pass to waveform-data.js.

I also wrapped the `peaks.ready` event in a 0ms `setTimeout`. This is because there was a race condition when supplying `rawData` as our source. Since we were no longer waiting for the response, the event was firing before the listeners could attach.


is343:
I see that it's failed two tests. I changed the error text because there are now three options to build from, and the text is not matching what the test is looking for.

Would you like me to update the test and/or add some new ones checking the validity of the raw data sources?

is343:
@chrisn I think I may have introduced a bug causing the segment markers for non-editable segments to be visible. Please see #309.

I have no idea how or why this this is happening. Do you have any ideas on what I might have changed that would cause this?

chrisn:
Thanks! This is a really useful feature. I have made a couple of small changes, and fixed the test cases, so please update your branch.

is343:
Thanks! I have pulled the changes.

is343:
I'm still getting that issue that I mentioned earlier. Are you able to test it to see if it's just me?

I tested on my end my running the build command to generate the peaks.js file and copied the contents into the one in my node_modules folder. So, this could have also been causing an issue.

chrisn:
The segment issue is unrelated, I'll push a fix later.

is343:
Alright, thanks! Let me know if you need me to do anything else!

README.md
```diff
@@ -478,12 +478,15 @@ peaks.on('ready', function() {
 
 Changes the audio or video media source associated with the `Peaks` instance. Depending on the `options` specified, the waveform is either requested from a server or is generated by the browser using the Web Audio API.
 
-The `options` parameter is an object with the following keys. Either `dataUri` or `webAudio` must be specified, but not both.
+The `options` parameter is an object with the following keys. Only one of `dataUri`, or `rawData`, or `webAudio` must be specified.
```
chrisn: I might want to think about the naming here, maybe `waveformData`? But then I'd also want to consider whether the other options (in particular, `dataUri`), should be reorganised.

peaks.js.d.ts
```diff
@@ -45,41 +45,50 @@ declare module 'peaks.js' {
   }
 
   interface RequiredOptions {
-    // HTML5 Media element containing an audio track
+    /** HTML5 Media element containing an audio track */
```
chrisn: Why change the comment format - is there tooling that makes use of this? I'm concerned about having to maintain the documentation in multiple places (here and in the ReadMe)

src/waveform-builder.js
```diff
@@ -96,6 +108,9 @@
     if (options.dataUri) {
       return this._getRemoteWaveformData(options, callback);
     }
+    else if (options.rawData) {
```
chrisn: We should add some test cases (test/api-spec.js). And (not essential) but we could consider adding or changing one of the demo pages to show how to use this feature.

peaks.js.d.ts
```diff
@@ -290,14 +309,24 @@ declare module 'peaks.js' {
...
-    // Events
+    /** Events */
     on: <E extends keyof InstanceEvents>(event: E, listener: InstanceEvents[E]) => void;
   }
 
   type PeaksInitCallback = (error: Error, peaks?: PeaksInstance) => void;
 
   interface PeaksOptionsWithoutAudioOptions extends RequiredOptions, OptionalOptions {}
 
+  export interface JsonPeaks {
```
chrisn: Suggest renaming to `JsonWaveformData`.

README.md
is343: How about `waveformData` and `waveformUri`?

peaks.js.d.ts
is343: Some editors (vscode) support syntax highlighting out of the box. [TSDoc](https://github.com/Microsoft/tsdoc) is similar to JSDoc.

The comments aren't necessary, I just figured since they were already there that I might as well change them.

I took a few screenshots as examples of what I see in vscode.

Screen Shot 2020-02-05 at 20 06 45
Screen Shot 2020-02-05 at 20 06 59
Screen Shot 2020-02-05 at 20 07 08
Screen Shot 2020-02-05 at 20 09 17
Screen Shot 2020-02-05 at 20 09 24
Screen Shot 2020-02-05 at 20 09 37

peaks.js.d.ts
is343: I agree. It's a bit more descriptive.

src/waveform-builder.js
is343: Sure, I can try adding some tests using the test data.

README.md
is343: We could also maybe just make a single `waveform` or a `preRendered` object and nest `data` and `uri` inside. Or perhaps `local` and `remote`.

is343: Or perhaps `remoteWaveform` and `localWaveform`.
How do you think you would like to reorganize things?

chrisn: For now, let's go with `waveformData` instead of `rawData`, for now. I'm thinking of simply renaming `dataUri` to `waveformDataUrl`, but let's do that separately to this PR.

src/waveform-builder.js
chrisn: Thanks!

is343: tests created!

test/unit/waveform-builder-spec.js
```diff
@@ -193,6 +193,60 @@ describe('WaveformBuilder', function() {
       }).to.throw();
     });
 
+    ('ArrayBuffer' in window) && it('should throw if arraybuffer data is invalid', function(done) {
+      expect(function() {
```
chrisn: It seems strange for a function that takes a callback to throw errors. I realise you're following the existing convention in the codebase, which is good, but this is something that should be fixed, really. I'll take a look.

is343: You're right, I should have made some for `Peaks.init` as well. 
I see that you've changed the tests a bit. I will add those tests following your new format tomorrow.

chrisn: Thank you! I have made a change so that callbacks are used consistently for error handling, so please update your branch again.

is343: New tests created! Please let me know if anything else needs to be changed!

is343:
Nope, no changes. You can go ahead and merge.

chrisn:
@is343 I've just published v0.20.0, which includes this update. Thanks again!

Issue #296, opened 2019-12-03
Allow different Audio Player (i.e. Tone.js) to work with Peaks.js Waveform


tscz:
Hello @chrisn,

this is more a feature request than an issue: It would be great if the peaks.js waveform could be linked with another player/library like Tone.js or Howler.js.

Currently the peaks.js waveform is tightly coupled with the <audio>-webaudio player. Out-of-the-box it is not possible to link the waveform with an external player (like Tone.js or Howler.js).

However, it is possible to replace the player with a compatible class but it comes with a price of code duplication, unofficial patching and using internal api (peaks.emit()). Very fragile, see https://github.com/tscz/transcribe/blob/ae85033b5edf768e60123e2b90cbf2894b36e4a3/src/features/audio/audioPlayer.ts#L20.

Is there any interest supporting this feature? If so, I would love to contribute here. 

Thx and regards
Teje

tscz:
FYI: My usecase is that I want to use Peaks.js as the Waveform Viewer and Tone.js as a player for adding a metronome or drumbeat to the playback and changing pitch and playbackrate.

chrisn:
I'm open to the idea of supporting Web Audio generated audio, as an option in addition to binding to a media element. I'd want to avoid taking a dependency on any specific Web Audio library though, so is there a way to do this using purely `AudioContext`? And I'd also want to avoid any breaking changes to the Peaks.js API for existing users.

tscz:
I added a design sketch in https://github.com/tscz/peaks.js/commit/2f7bdf0c9ec30223885cdaf4b0bd7e5f224f3f9c.

The implentation idea is to split the Player in a general part, which can be extended by an adapter pattern and a default implementation based on html5 media element. Replacing the default player then boils down to define an adapter.
```javascript
var adapter = {
    isReady: function() {},
    destroy: function() {},
    play: function() {},
    pause: function() {},
    isPlaying: function() {},
    isSeeking: function() {},
    getCurrentTime: function() {},
    getDuration: function() {},
    seek: function(time) {},
    playSegment: function(segment) {}
}

CustomPlayer = new Player(peaks,adapter);
```
No dependency to another library, not even a dependency to the AudioContext. And also no breaking changes to the Peaks.js API for users. Is this going into the right direction for you?

alittlebitweird:
I'm also very interested in this!

chrisn:
[...]

PR #314, opened 2020-02-29
Enable external players to be used with Peaks.js

tscz:
With these changes it is possible to use external players with Peaks.js. A configuration option.player is added. 

See https://github.com/bbc/peaks.js/issues/296.


tscz:
Hi @chrisn, I have not added documentation yet, maybe can you tell me where it should be added? Would you like further infos in README.md and/or in customizing.md?

chrisn:
Thanks, I'd like to try this out, do you have an example player adapter I could use?

I suggest adding the detail to customizing.md, with a short section in README.md like [this one](https://github.com/bbc/peaks.js/blob/master/README.md#marker-customization).

tscz:
@chrisn: I added a demo page based on tone.js to this merge request (https://github.com/bbc/peaks.js/pull/314/commits/c406070abd7ce768404c2adbcb94a91963b82f3a). Needs more polishing, but that might be a good example for you.

chrisn:
I've just had a quick look at the code. Before doing a deeper review, I noticed that the indentation is inconsistent in many places. We use a 2-space indent, but there isn't an eslint rule checking for this right now. Would you mind fixing this please, or let me know if you'd prefer me to do it.

tscz:
@chrisn: Good point, I fixed the inconsistent code formats by adding eslint rules for it and used the autofix. However, I didn't want to mix this pull request with reformatting unrelated code, so I only changed the classes I created here. 

I think enabling this eslint rules would be a good idea in general, so I created an isolated pull request https://github.com/bbc/peaks.js/pull/315.

tscz:
@chrisn: Added documentation and updated demo page. Ready for first review.

chrisn:
Is it possible to refactor to avoid using inheritance? e.g., I'm thinking in main.js we could have:

```js
if (instance.options.player) {
  instance.player = new Player(
    instance,
    instance.options.player
  );
}
else {
  instance.player = new Player(
    instance,
    new MediaElementPlayer(instance.options.mediaElement)
  );
}
```

Also, I tried to push a small change to your branch, but wasn't allowed. Would you prefer I write suggestions here?

chrisn:
Another question, about Tone.js in particular: can I use Peaks with non-`AudioBuffer` content, e.g., synthesized sounds?

tscz:
@chrisn: Sry, forgot to give you access to the fork repo. Feel free to push changes into the branch, perfectly fine with me.
. 
Regarding inheritance: Sure, it should be possible to get rid of it. I will have a look.

Regarding Tone.js: Yes and No;-) This library is based on a virtual timeline abstraction called [Transport](https://github.com/Tonejs/Tone.js/wiki/Transport). Events (like scheduling synthesized sounds or playing samples from an audio file) are registered at this Transport and played based on the position in this time line. Directly using it as direct input for Peaks is hence not possible, because it is basically only a timeline with scheduled events which dynamically creates a sound at point in time. However, there is an option to pre-compile the Transport into an AudioBuffer, see https://tonejs.github.io/examples/offline.html. This would then allow to display the AudioBuffer in Peaks.js, but you need to compile it after each change of the Transport structure and re-init Peaks' Waveform.

But by enabling to change the audio player allows to combine both worlds: My usecase for example is to use an mp3 song file as base and synthesized sounds (mentronome klick, drum beat patterns) on top of it. Hence the AudioBuffer of the song is input for Peaks, whereas the the synthesized sounds are created as implementation details of the Tone.js during Playback. The waveform is representing the mp3 file but seeking and play/pause via Peaks is basically controlling the Transport Object of Tone.js.

Hope this answers your question regarding usage of non-AudioBuffer content with Peaks?

chrisn:
This makes sense, thank you for explaining! I have pushed changes to the demo page to your branch.

tscz:
@chrisn: ready for second review: I removed Inheritance and made the refactoring in main.js as you suggested in https://github.com/bbc/peaks.js/pull/314#issuecomment-602498663. It comes with one trade-off:
* The `setSource `and `getCurrentSource `methods make only sense in the context of a `MediaElementPlayer` adapter, however they needed to be included in the `Player` class, since they are used within Peaks internally. These methods are thus no public player adapter API, and therefore I have hidden them in the TS type description and labeled `_setSource `and `_getCurrentSource `in the player.

tscz:
>     * The `setSource `and `getCurrentSource `methods make only sense in the context of a `MediaElementPlayer` adapter, however they needed to be included in the `Player` class, since they are used within Peaks internally. These methods are thus no public player adapter API, and therefore I have hidden them in the TS type description and labeled `_setSource `and `_getCurrentSource `in the player.

Having a second look, it is quite strange that the player is in charge for `setSource ` and `getCurrentSource ` in the first place. The options.mediaElement itself should be controling the source. I have to have a further look in the usage of these methods.

tscz:
@chrisn: Removed the setSource and getCurrentSourc functionality completely from the player. Peaks used these player functions in two contexts
1.) main.js in Peaks.setSource(). As last step the src property of the media element must be changed. This is however independent of the player (in general) and must be done on the media element directly.
2.) waveform-builder.js in WaveformBuilder._buildWaveformDataUsingWebAudio(). Within this function ther is a fix to deal with a race condition which used the player to read the currentSrc. Also this is (in general) a mediaelement property independent of the player impl. It is sufficient to register the callback on the "player_canplay" to deal with the race condition.

IMHO now the seperation of concerns for the player looks way better and the removal of inheritance was really a good idea:-)

chrisn:
Thank you for all your efforts on this. I should have some time to review, so I'll come back to you soon.

src/player.js
```diff
@@ -1,219 +1,180 @@
...
-
-  /**
+    /**
    * Starts playback.
    */
 
-  Player.prototype.play = function() {
-    this._mediaElement.play();
-  };
+    self.play = function() {
```
chrisn, line 70: I'd prefer to continue to use the `Player.prototype.play = function()` style, here and in MediaElementPlayer, if we can.

demo/external-player.html
```diff
@@ -0,0 +1,256 @@
...
+                Tone.connectSeries(externalPlayer, Tone.Master);
+
+                Tone.Transport.scheduleRepeat(time => {
+                  internalPlayer._updatedTime();
+
+                  if (internalPlayer.getCurrentTime() >= internalPlayer.getDuration()) {
+                    Tone.Transport.stop();
+                    onSongComplete();
+                  }
+                }, 0.03);
```
chrisn, line 167: Does this need to be faster than the media element's default, which is about every 250 milliseconds? Depending on zoom leve, Peaks.js will switch to using `requestAnimationFrame` and polling the player's current time.

src/player-mediaelement.js
```diff
@@ -0,0 +1,143 @@
...
+        self._player._triggeredCanPlay();
+      });
+
+      self._addMediaListener('error', function(event) {
+        self._player._triggeredError(event.target.error);
+      });
+
+      self._interval = null;
+    };
+    /**
```
chrisn, line 76: Minor nit: indentation

peaks.js.d.ts
```diff
@@ -250,6 +252,23 @@ declare module 'peaks.js' {
...
+    isPlaying: () => boolean;
+    isSeeking: () => boolean;
+    getCurrentTime: () => number;
+    getDuration: () => number;
+    seek: (time: number) => void;
+    playSegment: (segment: Segment) => void;
+  }
+
+  interface InternalPlayer extends PlayerAdapter {
+    _updatedTime: () => void;
```
chrisn, line 269: I'd prefer to name this `_timeUpdate`.

src/player.js
tscz: Done, switched to prototype style for Player and MediaElementPlayer.
As a consequence the adapter validation must be done on the adapter object and its prototype. Added a `getAllPropertiesFrom(adapter)` to this purpose.

src/player-mediaelement.js
tscz: Done.

peaks.js.d.ts
tscz: Done, renamed it to _timeUpdate.

demo/external-player.html
tscz: You're right, since this `Tone.Transport.scheduleRepeat()` is only used to inform about the current time, the 250ms default is fine. A user might also add want to add synth events at different intervals, but he will choose a different `Tone.Transport.scheduleRepeat()` then. Changed the example accordingly.

chrisn:
This looks great, thank you! I'll pick up from here and will merge soon.

elastisteve:
This is great! Can't wait to use it with Howler and Web Audio. Hope it all plays nice together...

chrisn:
@tscz I've found that some of the player events aren't being emitted at the right time (e.g., `player_seek` is now emitted when seek starts, not when it ends). We'll need to move responsibility for emitting these events into the custom player and `HTMLMediaElementPlayer` objects and adjust the tests.

tscz:
@chrisn : Mmh, true, I'll have a look. The events were emitted earlier exclusively after the media element events were fired, the current impl is not respecting this. Even worse, seek() invocation is emitting "player_seek" twice now, once at the start (via sync incovation in `Player.prototype.seek()`) and once at the end (via async invocation in `addMediaListener('seeked',function() {...})`

The current player events and corresponding media elements are

1. `player_time_update` and `timeupdate` (Fired when the time indicated by the currentTime attribute has been updated)
2. `player_play` and `play` (Fired when the paused property is changed from true to false)
3. `player_pause` and `pause` (Fired when a request to pause play is handled and the activity has entered its paused state)
4. `player_canplay` and `canplay` (Fired when the user agent can play the media, but estimates that not enough data has been loaded to play the media up to its end without having to stop for further buffering of content)
5. `player_error` and `error `(Fired when the resource could not be loaded due to an error)
6. `player_seek` and `seeked` (Fired when a seek operation completes)

chrisn:
Thanks! I'm working through it now, I'll update the PR.

chrisn:
I have updated the code to move the event triggering to the custom player or the media element player. This isn't quite as clean as I'd like - for example, we have to emit `player_time_update` after a seek, to match the behaviour of the media element player.

I also changed the `mediaElement` option, which shouldn't be required if you're using an external player (your Tone.js example doesn't need it). This needs some test cases adding.

tscz:
@chrisn: Short update: I triedadding missing testcases and stumbled upon issues with the media element player in the test environment.

I saw that the MediaElementPlayer.play() was not firing the 'play' event of the media element in the tests.
The play() function in Peaks.js is a void function, the MediaElement however returns a Promise. Investigating the promise like
```javascript
    describe('play', function() {
      it('should trigger mediaelement play event', function(done) {
        document.querySelector('audio').play()
          .then(function() {
            done();
          },function(reason) {
            console.log('code = ' + reason.code);
            console.log('message = ' + reason.message);
            console.log('code = ' + reason.name);
            done(new Error('Play did not succed'));
          });
      },false);
    });
```
produces the error message
```
LOG LOG: 'code = 0'
LOG LOG: 'message = play() can only be initiated by a user gesture.'
LOG LOG: 'code = NotAllowedError'
    with media element player
      play
        × should trigger mediaelement play event
        Error: Play did not succed
```
So actually using `play()` in the tests never worked, because the autoplay policy of headless chrome is preventing it. The promise always failed undetected.

During usage in a browser, I found general recommendations to invoke the play by a user interaction, like a button click. Simulating this in a unit test didn't work, why I do not know (yet).
Since we are here in a headless chrome test environment I think it is appropriate to disable this security mechanism and configured the karma.conf.js accordingly with a Browser profile
```
      'ChromeHeadlessWithoutAutoplayPolicy': {
        base: 'ChromeHeadless',
        flags: ['--autoplay-policy=no-user-gesture-required']
      }
```
Then the aforementioned smoke tests succeeds. However, since now play() is correctly causing media element events, the real production code side-effects are triggered. In concrete, the CueEmitter
```javascript
  CueEmitter.prototype._attachEventHandlers = function() {
    this._peaks.on('player_time_update', this._onTimeUpdate);
    this._peaks.on('player_play', this._onPlay);
    this._peaks.on('player_seek', this._onSeek);
```
registeres to the play event and causes the test to fail, because it wants to interact with the media element even though the test is already done. So I am in the dilemma of wanting to mock the emit('player_play') in the test to avoid side effects but needing it to wait in the async test via
```javascript
        p.on('player_play', function(currentTime) {
          done();
        });
```
Do you have an idea here?

I further investigate and need to understand the side effects of the player events.

chrisn:
I agree, disabling autoplay restriction makes sense in the tests.

I found that the following test isn't working. It hangs waiting for the event. But, if I run the test using `describe.only()`, it passes.

```js
describe('play', function() {
  it('should trigger a player_play event', function(done) {
    p.on('player_play', function() {
      done();
    });

    document.querySelector('audio').play();
  });
});
```


So I haven't been able to replicate your issue with CueEmitter yet.

tscz:
I pushed a solution proposal.

The side effect is, that the Cue emitter is during playback in an requestAnimationFrame recursion
```javascript
  CueEmitter.prototype.onAnimationFrame = function() {
    var time = this._peaks.player.getCurrentTime();
...
    if (this._peaks.player.isPlaying()) {
      this._rAFHandle = requestAnimationFrame(this._onAnimationFrame);
    }
  };
```
Thus it stays in this recursion, if the player is not pausing. After the test the player object is gone and the ` var time = this._peaks.player.getCurrentTime();` causes the "Uncaught TypeError: Cannot read property 'currentTime' of null".

Hence I added a pause() invocation in the unit test of the play() function to stop the recursive calls.
What do you think?

Fun fact: I also muted the media element, since now play() is playing real audio, which hit me. My speakers were loud, never have been startled by a unit test run before;-).

chrisn:
> Fun fact: I also muted the media element, since now play() is playing real audio, which hit me. My speakers were loud, never have been startled by a unit test run before;-).

Oh no! :laughing: 

I'm not sure the `CueEmitter` is the issue, because that's only created if the `emitCueEvents` option is set, so it shouldn't be running outside if its own unit tests.

But, there's another rAF loop, inside the `Konva.Animation` in `PlayheadLayer`, which I found to start running during the player tests. So I think you've uncovered a bug ... :+1: The animation needs to be stopped when `peaks.destroy()` is called. We also need to ensure all tests have an `afterEach` function that calls `peaks.destroy()`.
 
I'll push a fix for this.

tscz:
Perfect, you are absolutely right! I draw the wrong conclusion from the stacktrace and related it wrongly to the `CueEmitter `and not the `Konva.Animation`.

One question: I want to define the interface for the stub player in the Typescript Definition file. After the changes now, how is the public api for the parameter in `player.init(?)`?

Is it the complete Peaks Instance and the "player_" events are added to the public `InstanceEvents`? Or are the "player_" events still internal and only in this context of the `init` method the parameter would be an `EventEmitter` who knows about the "player_" events?

chrisn:
I'd like your feedback on the custom player API. Is it OK that the custom player emits events as per the media element, so `init()` only needs an event emitter, or do we still need an API between the custom player and internal player as before?

I'm thinking of making the player events public - see https://github.com/bbc/peaks.js/issues/318#issuecomment-617106422. I was planning to do this as a separate change.

tscz:
I would vote for an event emitter as input parameter. That is sufficient for custom player implementations, which will sort of all have the same basic issues: wrap basic player functions and inform Peaks.js about invocations/updates. 

My proposal in total for the public API would be:
```typescript
  interface PlayerAdapter {
    init: (eventEmitter: EventEmitterForPlayerEvents) => void;
    destroy: () => void;
    play: () => Promise<void>;
    pause: () => void;
    isPlaying: () => boolean;
    isSeeking: () => boolean;
    getCurrentTime: () => number;
    getDuration: () => number;
    seek: (time: number) => void;
    playSegment: (segment: Segment) => void;
  }

  interface EventEmitterForPlayerEvents {
    on: <E extends keyof PlayerEvents>(event: E, listener: PlayerEvents[E]) => void;
  }

  interface PlayerEvents {
    player_canplay: () => void;
    player_error: (error: any) => void
    player_pause: (time: number) => void;
    player_play: (time: number) => void;
    player_seek: (time: number) => void;
    player_time_update: (time: number) => void;
  }
````

chrisn:
This looks good, but I'd suggest one minor change: `play()` may also return `void`, for legacy browsers that don't return a promise. I'll push a change to the tests for this.

chrisn:
Would you like to update the TypeScript definitions and documentation?

After this, I'll probably make some edits to the documentation, and then I think this will be ready to merge.

tscz:
Sure, will do so.

tscz:
I pushed the typescript definitions and updated the documentation.

chrisn:
Thanks! I made one change to the API - I've moved `playSegment` to the`Player`, as this can be implemented in the same way for the `HTMLMediaElementPlayer` and custom players.

I've also made a lot of editorial changes to the documentation. I hope this is OK with you.

Would you please check that everything works as you expect, then i think we can go ahead and merge?

tscz:
@chrisn: Checked it. Adjusted the unit test for the playSegment, build is green. Also added a showcase to the example page. The documentation reads very good, with the examples per method and event is way clearer than my initial table approach!

Double checked it from my perspective as a future user and everything is there. From my side you can go ahead with the merge.

chrisn:
I should have fixed the tests myself, sorry!

chrisn:
This all looks good to me, so will merge now. Thank you so much for contributing! :pray:

tscz:
This is great! Thanks a lot for your support of this feature and all the effort to get this production ready, I can assure you have already one happy user of this new api:-)

chrisn:
It's a pleasure! Release 0.21.0 is now available on npm.

PR #315, opened 2020-03-15
Add eslint formatting indent rules

tscz:
This pull request adds eslint indent rules and enables the use of them during the linting build process.

The rules are derived from the current codebase. I tried to keep the formatting changes as small as possible.


chrisn:
This is great, thank you!

PR #319, opened 2020-04-21
Add option to set precision points in displayed timecode

Dananji:
This PR adds an option to Peaks.js to set number of precision points when displaying time in the Waveform.
In the current version, the default time code format is as `hh:mm:ss.ss`. With the addition of this new option, `precisionPoints` the user of the framework can specify the number of places the fractional seconds are calculated.
E.g. Doing the following;
```
let peaksInstance = Peaks.init({
  ...options,
  precisionPoints: 3
})
``` 
will result in time codes in `hh:mm:ss.sss` format like below;

Peaks js_precitionPoints

### DIFF
--- a/src/playhead-layer.js
+++ b/src/playhead-layer.js
@@ -185,7 +185,7 @@ define([
       this._playheadGroup.setAttr('x', playheadX);
 
       if (this._playheadText) {
-        var text = Utils.formatTime(time, false, precision);
+        var text = Utils.formatTime(time, precision);
 
         this._playheadText.setText(text);
       }
@@ -276,6 +276,26 @@ define([
         this._playheadText = null;
         updated = true;
       }
+    }
+
+    if (updated) {
+      this._playheadLayer.draw();
+    }
+  };
+
+  PlayheadLayer.prototype.setPlayheadTimePrecision = function(precision) {
+    var updated = false;
+
+    if (this._playheadText) {
+      // Set option in peaks
+      this._peaks.options.playheadTimePrecision = precision;
+
+      var time = this._peaks.player.getCurrentTime();
+      var text = Utils.formatTime(time, precision);
+
+      // Update current play head
+      this._playheadText.setText(text);
+      updated = true;
     }
 
     if (updated) {
--- a/src/playhead-layer.js
+++ b/src/playhead-layer.js
@@ -172,7 +172,7 @@ define([
                     (pixelIndex <  frameOffset + width);
 
     this._playheadPixel = pixelIndex;
-    var precision = this._peaks.options.precisionPoints;
+    var precision = this._peaks.options.playheadTimePrecision;
 
     if (isVisible) {
       var playheadX = this._playheadPixel - frameOffset;
--- a/src/playhead-layer.js
+++ b/src/playhead-layer.js
@@ -281,7 +281,7 @@ define([
     }
   };
 
-  PlayheadLayer.prototype.setTimeLabelPrecision = function() {
+  PlayheadLayer.prototype.updatePlayheadText = function() {
     // Update current play head
     if (this._playheadText) {
       var time = this._peaks.player.getCurrentTime();
--- a/src/playhead-layer.js
+++ b/src/playhead-layer.js
@@ -7,9 +7,8 @@
  */
 
 define([
-  './utils',
   'konva'
-], function(Utils, Konva) {
+], function(Konva) {
   'use strict';
 
   /**
@@ -172,7 +171,6 @@ define([
                     (pixelIndex <  frameOffset + width);
 
     this._playheadPixel = pixelIndex;
-    var precision = this._peaks.options.playheadTimePrecision;
 
     if (isVisible) {
       var playheadX = this._playheadPixel - frameOffset;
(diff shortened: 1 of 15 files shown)


chrisn:
Wow, thanks! Let me think about the naming / API to this feature.... I have one question in the mean time: What happens if I set the value to zero? I'd expect no fraction part to be shown.

Dananji:
> Wow, thanks! Let me think about the naming / API to this feature.... I have one question in the mean time: What happens if I set the value to zero? I'd expect no fraction part to be shown.

Right, I did not account for that. Thanks for catching that. I'll fix it :+1:

chrisn:
Regarding API, I've recently been adding methods to allow certain properties to be updated programmatically (see the [View API documentation](https://github.com/bbc/peaks.js#view-api)).

So we could add a method such as `view.setPlayheadTimePrecision()`. This allows people to customise the overview and zoomview independently if they want to, and allows the setting to be changed at any time, not just on construction.

Actually, I would prefer to do this than adding more and more options to `Peaks.init()`, but as yet we don't have another good way to set the initial value, so without an option, calling `Peaks.init()` then `view.setPlayheadTimePrecision()` would cause the view to redraw.

If we really do need an option, I would probably prefer to name it `playheadTimePrecision`.

What do you think?

Dananji:
Yes I agree. Having a method that we could use to change precision is a good solution when we want to change it after initializing the Peaks instance :+1: 
If there's no other way to set it at the time of initializing Peaks, then the name `playheadTimePrecision` for this options sounds good to me.

chrisn:
OK, let's go with `playheadTimePrecision`, and also add `view.setPlayheadTimePrecision()` (which would be similar to the existing `view.showPlayheadTime()`).

We can think about the other issues I mentioned another time, as they'll probably introduce bigger changes to the API.

Dananji:
Should I add `view.setPlayheadTimePrecision()` in PR as well, since its related?

chrisn:
I leave that up to you ... I can work on it later, if you prefer.

Dananji:
I'd like to try, if I can't make it work then I'll leave it to you.

Dananji:
@chrisn I added the `view.setPlayheadTimePrecision()` method.

src/waveform-overview.js
```diff
@@ -344,6 +344,10 @@ define([
     this._playheadLayer.showPlayheadTime(show);
   };
 
+  WaveformOverview.prototype.setPlayheadTimePrecision = function() {
+    // The overview waveform doesn't show time codes for play head and markers
```
chrisn, line 348: This method can apply to the overview waveform, as the user can call `showPlayheadTime(true)`.

Dananji: Oh okay. I'll change it along with the README.

chrisn:
Looks good, thank you. I think it's ready to merge, unless you have more changes to add?

chrisn:
Sorry, I've just found a couple of small issues:

* If you change the precision while the time label isn't shown, the updated precision isn't applied.
* I generally avoid changing the options object, preferring to copy values into internal state (for example, see how the `playheadColor` option is handled [here](https://github.com/bbc/peaks.js/blob/f757fdcfb8fb0d30afb48ca7e40a74fd06f6a251/src/playhead-layer.js#L33))

Dananji:
For the second issue, the question I have is how we would get the new time precision value in segments and points? Should we introduce another method to get the internal state `this._playheadTimePrecision` from `playhead-layer.js` into segments and points?

chrisn:
I see, I hadn't spotted that this change applies to the both playhead and the point/segment markers. Sorry.

In that case, `playheadTimePrecision` isn't the right name, perhaps `timeLabelPrecision` instead?

Regarding getting the precision to the points and segments, we could move responsibility for formatting the time string from `DefaultPointMarker` and `DefaultSegmentMarker` to the layer or view classes, e.g:

```js
DefaultPointMarker.prototype.timeUpdated = function(time) {
  if (this._time) {
    this._time.setText(this._options.layer.formatTime(time));
  }
};

PointsLayer.prototype.formatTime = function(time) {
  // Either:
  return Utils.formatTime(time, this._timeLabelPrecision);
  // or:
  return this._view.formatTime(time);
}

WaveformOverview.prototype.formatTime = function(time) {
  return Utils.formatTime(time, this._timeLabelPrecision);
}
```

If we move the responsibility to the view, then we can also call this from `PlayheadLayer`. It would also mean that we don't need to pass the option through the marker factory functions.

What do you think?

Dananji:
Yes, `timeLabelPrecision` sounds much better :+1: 
Moving formatting time responsibility to view classes feels like the best way to do it. So that we a single point in the code which responds to the time precision changes.

src/playhead-layer.js
```diff
@@ -282,5 +281,16 @@ define([
     }
   };
 
+  PlayheadLayer.prototype.setTimeLabelPrecision = function() {
```
chrisn, line 284: As we're not passing a new precision value to this function, i'd suggest renaming this as `updatePlayheadText()`

chrisn:
I'll merge now. Thank you for contributing, and for your patience working through the details.

Dananji:
Thanks for spending time to go through this and helping me.

chrisn:
I've just published v0.20.0, which includes this feature. Thanks again!

PR #322, opened 2020-06-23
put playheadText left if playhead is in right end

Karoid:
AC_ 20200623-142503

This code put text on left side when playheadText can be hidden.


chrisn:
Thank you! This is a good idea. I notice there are a few linter warnings (`npm run lint`), could you please fix these?

Karoid:
I fixed eslint warnings and errors

chrisn:
Thank you for contributing, we really appreciate it!

Issue #351, opened 2020-12-20
Using setSource with audio context and buffer


Christilut:
Currently I'm loading audio into Peaks through `init` as such:
```ts
    const audioContext: AudioContext = new AudioContext()
    const arrayBuffer = new Uint8Array(await readFile(location)).buffer
    const dataUri = 'data:' + getHowlerMimetype(location) + ';base64,' + base64ArrayBuffer(arrayBuffer)

    const audioBuffer: AudioBuffer = await audioContext.decodeAudioData(arrayBuffer)

 this.peaks = await Peaks.init({
      containers: {
        overview: this.overviewContainer,
        zoomview: this.container
      },
      mediaElement: undefined as any,
      webAudio: {
        audioContext,
        audioBuffer
      }
})
```
Full config in #349 (using Howler as player).

This works but I'd like to re-use Peaks without doing a re-init when I change audio file.

According to the documentation I can use setSource so I tried this:
```ts
      this.peaks.setSource({
        webAudio: {
          audioContext,
          audioBuffer
        }
      }, () => {
        console.log('loaded')
      })
```

The callback fires but no waveform is loaded in Peaks.

chrisn:
Does your player emit a `player.canplay` event?

Christilut:
Yes, this is my player config:
```ts
      player: {
        init: (eventEmitter) => {
          this.eventEmitter = eventEmitter

          this.eventEmitter.emit('player.canplay')

          setInterval(() => {
            const time = this.getCurrentTime()

            eventEmitter.emit('player.timeupdate', time)

            if (time >= this.getDuration()) {
              this.howl.stop()
            }
          }, 0.25)
        },

        destroy: () => {
          this.howl?.unload()
        },

        play: () => {
          this.howl.play()
          this.playing = true

          this.eventEmitter.emit('player.play', this.getCurrentTime())
        },

        pause: () => {
          this.howl.pause()
          this.playing = false

          this.eventEmitter.emit('player.pause', this.getCurrentTime())
        },

        seek: (seconds) => {
          this.howl.seek(seconds)

          this.eventEmitter.emit('player.seeked', this.getCurrentTime())
          this.eventEmitter.emit('player.timeupdate', this.getCurrentTime())
        },

        isPlaying: () => {
          return this.howl.playing()
        },

        isSeeking: () => {
          return false
        },

        getCurrentTime: () => {
          return this.getCurrentTime()
        },

        getDuration: () => {
          return this.getDuration()
        }
[...]

PR #352, opened 2020-12-22
setSource mediaUrl optional

Christilut:
Removed the required for mediaUrl in the setSource method.

If no mediaUrl is present, it will run `playerCanPlayHandler` since no callback is made otherwise.

See #351

### DIFF
--- a/peaks.js.d.ts
+++ b/peaks.js.d.ts
@@ -48,10 +48,7 @@ declare module 'peaks.js' {
     update: (options: PointUpdateOptions) => void;
   }
 
-  interface RequiredOptions {
-    /** HTML5 Media element containing an audio track */
-    mediaElement: Element;
-  }
+  interface RequiredOptions {  }
 
   interface SingleContainerOptions {
     /** Container element for the waveform views */
@@ -158,6 +155,8 @@ declare module 'peaks.js' {
   }
 
   interface OptionalOptions {
+    /** HTML5 Media element containing an audio track. Optional when using an ArrayBuffer */
+    mediaElement?: Element;
     /**
      * If true, peaks will send credentials with all network requests
      * - i.e. when fetching waveform data.
--- a/src/main.js
+++ b/src/main.js
@@ -557,6 +557,13 @@ define([
   Peaks.prototype.setSource = function(options, callback) {
     var self = this;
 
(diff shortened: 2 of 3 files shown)


chrisn:
Thank you! We'll need to add some test cases around this. The `mediaUrl` should be optional only if you're using an external player object. If you're using the HTMLMediaElement based player then it should still be required.

Christilut:
I've changed the typings so that `mediaElement` is no longer required since it's not used when using an arraybuffer.

I'll add some tests soon.

chrisn:
Would you mind putting this change on a branch, please?

src/main.js
```diff
@@ -557,8 +557,9 @@ define([
   Peaks.prototype.setSource = function(options, callback) {
     var self = this;
 
-    if (!options.mediaUrl) {
-      callback(new Error('peaks.setSource(): options must contain a mediaUrl'));
+    if ((options.mediaElement || this.options.mediaElement) && !options.mediaUrl) {
```
chrisn, line 560: `mediaElement` can only be set in `Peaks.init()`. It shouldn't be a valid option in `setSource()`.

peaks.js.d.ts
```diff
@@ -48,10 +48,7 @@ declare module 'peaks.js' {
     update: (options: PointUpdateOptions) => void;
   }
 
-  interface RequiredOptions {
-    /** HTML5 Media element containing an audio track */
-    mediaElement: Element;
-  }
+  interface RequiredOptions {  }
```
chrisn, line 51: I think we should have made `mediaElement` optional when we added the `player` option, so thanks for catching this. As `RequiredOptions` is now empty, we should remove it entirely.

src/main.js
```diff
@@ -608,7 +609,12 @@ define([
     self.once('player.canplay', playerCanPlayHandler);
     self.once('player.error', playerErrorHandler);
 
-    self.options.mediaElement.setAttribute('src', options.mediaUrl);
+    if (options.mediaUrl) {
```
chrisn, line 612: I think this should check for `this.options.mediaElement`. It amounts to the same thing, but seems clearer to me (If there's a media element, change it's source and wait for a `canplay` event, if there's a custom player, create the waveform directly without waiting for an event).

PR #353, opened 2020-12-22
Added support for gradients

Christilut:
Added support for waveform gradient colors.

Limited to 2 color stops:
```
        overviewWaveformColor: {
          linearGradientStartPoint: 50,
          linearGradientEndPoint: 80,
          linearGradientColorStops: [0, 'hsl(120, 78%, 26%)', 1, 'red']
        },
```

image


chrisn:
Thanks for this, it's a nice feature. A few comments:

* What units are the `linearGradientStartPoint` and `linearGradientEndPoint`?
* If we allow gradients here, we should also enable it [`peaks.segments.add()`](https://github.com/bbc/peaks.js#instancesegmentsaddsegment) (There's a missing check for the segment color in `validateSegment`)
* Please use `Utils.isString` and `Utils.isObject` in place of the `typeof` checks
* Please update the TypeScript definitions file, peaks.js.d.ts
* Please revert the line endings change, which is unrelated to this feature
* We should add some unit tests...

Christilut:
Changed things based on your feedback.
The line endings commit is a revert, for some reason my vscode added the COPYING file without changes

I'll fix the tests and add more soon.

chrisn:
> What units are the linearGradientStartPoint and linearGradientEndPoint?

Specifying these points in pixels means that the colour gradient will change if the user resizes the waveform view, so to keep the gradient consistent the user will have to call `view.setWaveformColor()` after `view.fitToContainer()`, not only for the main waveform view but also for any segments using colours gradients. I'd prefer this to be handled automatically, e.g., by specifying the start and end points as percentage of the view height.

Christilut:
I've changed it to percentages.
Konva seems to handle resizing well, but I'm not sure if there are any other places that I need to explicitly do anything? I guess this is all done by Konva already as is

chrisn:
Thanks! The easiest way to test this is the Resize button in the demo page at demo/index.html. The code to handle resizing is in `WaveformZoomView.fitToContainer()`, `SegmentsLayer.fitToView()`, and `SegmentShape.fitToView()`.

Christilut:
Done! They resize properly now, including the segments.
I also added `test_data` to gitignore because it kept recreating that when launching the demo.

chrisn:
Sorry for the delay. This all looks good. One minor thing I'd like to change is that we don't need to add `Percentage` in the variable names, it's enough to state the units in the documentation.

Christilut:
Done!

chrisn:
Thanks a lot! I may make a couple of minor changes before merging, but this all looks good.

I'll reply to your other PRs soon, by the way.

Christilut:
Awesome, thanks :)

jgentes:
FYI this is in the documentation but I don't think it has been released yet (last npm release on Nov 27)

Thanks @Christilut for the work on the PR and @chrisn for the very impressive work on this module.

chrisn:
Good point, I'll make a new npm release soon.

jgentes:
FYI this still doesn't appear to be in the `0.24.1` release - from `peaks.js.d.ts`, `setWaveformColor` can only be a string:

`setWaveformColor: (color: string) => void;`

chrisn:
It's implemented, but the `setWaveformColor` TypeScript declaration hadn't been updated. Fixed in b957e4dd.

jgentes:
I can't seem to get the gradient working.. always just gives a green waveform, even using your example settings. I've setup a repro if you want to take a look: https://codepen.io/jgentes/pen/YzNjdpL

chrisn:
I see what's happening. If you set a gradient in `Peaks.init()` you can then use `setWaveformColor()` to change the gradient. If you set a solid colour in `Peaks.init()` then later calling `setWaveformColor()` with a gradient doesn't work.

chrisn:
This is now fixed in release v0.25.0. Note that I have also changed the `Peaks.init()` API because there's potential for race condition during initialisation. You'll need to pass a callback function rather than use the return value.

jgentes:
Thanks Chris! In my 'real' code, I am using the callback :)

PR #354, opened 2020-12-22
Playhead fixed center

Christilut:
This is a work in progress, just need some feedback before I can continue.

This PR adds the `playheadFixedCenter` option and keeps the center of the zoom waveform at the center.
I've tested it with the Points API and it works well.
The segments act strange though, they only seem to appear when the center is in the middle of a segment. Can't really figure out why, would like a pointer here if possible.

Another small thing I wasn't able to figure out yet: When a track is at 0 seconds, it should start at the center but now it still starts at the left side of the view. I suppose there needs to be `width / 2` padding on the left of some sort but I don't know how to do that here.

And the last thing is that the current position marker in the overview needs to be adjusted by `highlight width / 2` but I'm not quite sure where to get the highlight width. Is it available somewhere in the HighlightLayer or should I pass that around?

I'll add tests when everything is done.

### DIFF
--- a/peaks.js.d.ts
+++ b/peaks.js.d.ts
@@ -229,6 +229,8 @@ declare module 'peaks.js' {
     points?: Point[];
     /** Emit cue events when playing */
     emitCueEvents?: boolean;
+    /** Fixes the zoom waveform playhead marker at the center of the container. */
+    playheadFixedCenter?: boolean
     /** Custom segment marker factory function */
     createSegmentMarker?: (options: CreateSegmentMarkerOptions) => SegmentMarker;
     /** Custom segment label factory function */
--- a/src/main.js
+++ b/src/main.js
@@ -258,6 +258,11 @@ define([
       emitCueEvents: false,
 
       /**
+       * Fixes the zoom waveform playhead marker at the center of the container.
+       */
+      playheadFixedCenter: false,
+
+      /**
        * Point/Segment marker customisation.
        *
        * @todo This part of the API is not stable.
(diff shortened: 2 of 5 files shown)


chrisn:
> When a track is at 0 seconds, it should start at the center but now it still starts at the left side of the view.

What I'd probably do is allow `frameOffset` to go negative, which would need changes to [WaveformShape._sceneFunc](https://github.com/bbc/peaks.js/blob/6eed95322b6303c8a7c8fbff9d3a7c35e37bac0c/src/waveform-shape.js#L97) and/or [WaveformShape._drawWaveform](https://github.com/bbc/peaks.js/blob/6eed95322b6303c8a7c8fbff9d3a7c35e37bac0c/src/waveform-shape.js#L128).

For example, if you wanted to place time 0 in the waveform 100 pixels to the right, you'd set `frameOffset` to -100, then in `_drawWaveform`, you'd avoid drawing waveform points for any negative offset.

chrisn:
To find out why segments aren't being shown correctly, I added logging to [`SegmentsLayer.updateSegments()`](https://github.com/bbc/peaks.js/blob/6eed95322b6303c8a7c8fbff9d3a7c35e37bac0c/src/segments-layer.js#L175). It seems that the `startTime` and `endTime` values don't correspond to the time range that's visible in the view.

chrisn:
One small suggestion for this feature, perhaps the playhead position could be user-defined rather than fixed to the centre?

Christilut:
Thanks! I gave it all some more thought and I think my use case is too specific for peaks so better if I start from scratch. There's too many features that I need that would probably make peaks messy :)

PR #362, opened 2021-01-28
Add VSCode & WebStorm IDE folders to .gitignore

ffxsam:
Minor change to ignore config folders created by VS Code & WebStorm


chrisn:
This belongs more in your local .gitignore file, so it applies to all your projects. Otherwise we need to add entries for every possible editor or IDE that people use.

ffxsam:
👍

PR #367, opened 2021-02-11
Added player.ended event

ffxsam:
This is emitted when the audio element itself emits an "ended" event, as well as when a segment played by `playSegment()` reaches the endpoint.


chrisn:
This is a good addition, thank you!

ffxsam:
You're welcome! More coming soon. I may post in 'Discussions' soon so I can share some ideas and see what you think.

PR #369, opened 2021-02-20
Update README code example to use correct element ID

OmarShehata:
I copied the code snippet from the README while setting up PeakJS and was confused why I was getting an error. Turns out there are two code examples that incorrectly reference the HTML element IDs as `overview-waveform` and `zoomview-waveform` but they are set up in the very first example as `X-container` instead of `X-waveform`.

This PR updates these two examples to reference `overview-container` and `zoomview-container` to be consistent with all the other examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,8 +255,8 @@ const audioContext = new AudioContext();
 
 const options = {
   containers: {
-    overview: document.getElementById('overview-waveform'),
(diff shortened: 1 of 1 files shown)


chrisn:
Sorry that the documentation was misleading, and thank you for fixing it for us!

OmarShehata:
Thank you for the quick response and for maintaining this awesome library and making it freely available!

PR #393, opened 2021-06-07
Add root:true to eslintrc.js

jdelStrother:
Before this, when using peaks in a subdirectory of a larger project, peak's linting rules would combine with the parent project linting rules, which doesn't seem desirable.


chrisn:
Looks good, thank you!

PR #394, opened 2021-06-14
Fix overviewHighlightColor, update .d.ts

jdelStrother:
overviewHighlightColor stopped working in https://github.com/bbc/peaks.js/commit/93c7024ccfccfbfa13de2addac644492fb2219cf - this fixes that and updates the types to cover the new options structure.

(Update: actually, overviewHighlightColor continued working as long as you specified both it and the overviewHighlightOffset. But it was no longer applying the default values for either, and they couldn't be specified via the new 'options.overview' option.)

### DIFF
--- a/peaks.js.d.ts
+++ b/peaks.js.d.ts
@@ -62,7 +62,30 @@ declare module 'peaks.js' {
     }
   }
 
-  type ContainerOptions = SingleContainerOptions | ViewContainerOptions;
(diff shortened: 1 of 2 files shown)


chrisn:
Thanks, I missed the highlight options when I wrote this. Updating the TypeScript definitions was on my list of things to do (code on master branch may well be unfinished work in progress...), so thanks for adding them!

PR #400, opened 2021-06-22
[WIP] Fixing anchor links

rowild:
Work in progress

Some links don't work, because they are called e.g. "Marker methods" in the "[Table of] Contents" section, but actually refer to a headline called "Marker API". Here are some fixes.

Question: 
I fixed "Player methods" to "Player Interface", but maybe a consistent renaming to "Player API" would be more suitable?


rowild:
@chrisn This would need your review.

customizing.md
```diff
@@ -21,7 +21,7 @@ This document describes how to customize various aspects of the waveform renderi
 - [Segment Labels](#segment-labels)
   - [createSegmentLabel()](#createsegmentlabeloptions)
 - [Media Playback](#media-playback)
-  - [Player Methods](#player-methods)
+  - [Player Interface](#player-interface)
```
chrisn, line 24: Should we name this "Player API", to be consistent with "Marker API"?

rowild: It seems this was fixed during the recent work on the v0.27 release, right? If so, this PR could be deleted...

rowild: Not quite: the "Install with npm" in the TOC jump link does not work. The link should either be pointing to `#es-module-bundle` or the target headline be renamed to "Install with npm". New PR?

chrisn: I prefer to keep Player Interface here, as this describes the interface to external media players, whereas [Player API](https://github.com/bbc/peaks.js/#player-api) is part of Peaks.js public API.

Yes, let's make a new PR to fix the "Install with npm" link. The TOC entry should match the heading in the document, so we can either rename the heading to "Install with npm" or change the TOC entry to "ES module bundle" (and fix the link) - I don't mind which.

rowild:
Providing a new PR, so we can close this issue.

PR #402, opened 2021-06-22
Reduce frame offset wrangling

jdelStrother:
This is a bit of a grab-bag of commits broken out from my segment-dragging branch, which I think are useful whether or not we decide to merge segment-dragging.  I can break them out into separate PRs if you'd prefer - each commit is relatively independent so it might make more sense to review that way.

Do you consider methods like WaveformZoomview.prototype.pixelsToTime to be public? They have jsdoc comments and don't have a "_" prefix, but conversely they aren't in the main documentation or in the Typescript defs. If they're public I'll add them to peaks.d.ts, but I didn't want to add to the public API surface area more than necessary.

I also wanted to check about the removal of some `Math.floor`s - eg here: 

```diff
--- a/src/waveform-zoomview.js
+++ b/src/waveform-zoomview.js
@@ -150,11 +150,7 @@ define([
       onMouseUp: function(/* mousePosX */) {
         // Set playhead position only on click release, when not dragging.
         if (!self._mouseDragHandler.isDragging()) {
-          var mouseDownX = Math.floor(this.mouseDownX);
-
-          var pixelIndex = self._frameOffset + mouseDownX;
-
-          var time = self.pixelsToTime(pixelIndex);
+          var time = self.pixelOffsetToTime(this.mouseDownX);
           var duration = self._getDuration();

           // Prevent the playhead position from jumping by limiting click
```

pixelsToTime & pixelOffsetToTime are both going to multiply their argument by a floating point scale value so I couldn't see a good reason for trying to `floor` the original value to an integer. Let me know if there's something I'm missing, though.

### DIFF
--- a/peaks.js.d.ts
+++ b/peaks.js.d.ts
@@ -385,7 +385,8 @@ declare module 'peaks.js' {
     };
     /** Segments API */
     segments: {
-      add: (segments: SegmentAddOptions | SegmentAddOptions[]) => void;
+      add(segment: SegmentAddOptions): Segment;
+      add(segments: SegmentAddOptions[]): Segment[];
       getSegments: () => Segment[];
       getSegment: (id: string) => Segment | null;
       removeByTime: (startTime: number, endTime?: number) => Segment[];
@@ -394,7 +395,8 @@ declare module 'peaks.js' {
     };
     /** Points API */
     points: {
-      add: (points: PointAddOptions | PointAddOptions[]) => void;
+      add(point: PointAddOptions): Point;
+      add(points: PointAddOptions[]): Point[];
       getPoints: () => Point[];
       getPoint: (id: string) => Point | null;
       removeByTime: (time: number) => Point[];
--- a/peaks.js.d.ts
+++ b/peaks.js.d.ts
@@ -368,7 +368,7 @@ declare module 'peaks.js' {
     };
     /** Views API */
     views: {
-      createOverview: (container: HTMLElement) => WaveformOverView;
+      createOverview: (container: HTMLElement) => WaveformOverview;
       createZoomview: (container: HTMLElement) => WaveformZoomView;
       destroyOverview: () => void;
       destroyZoomview: () => void;
--- a/src/points-layer.js
+++ b/src/points-layer.js
@@ -83,11 +83,8 @@ define([
   };
 
   PointsLayer.prototype._onPointsUpdate = function(point) {
-    var frameOffset = this._view.getFrameOffset();
-    var width = this._view.getWidth();
-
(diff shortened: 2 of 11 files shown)


chrisn:
Thanks. I'll take a look a bit later...

The public interface is what's documented in the ReadMe and in customizing.md. Underscores are for things that are geniunely private (don't use from another class / module). Functions that don't have underscores but are also not in the ReadMe are for internal use only but can be called from other classes / modules - so wouldn't belong in the TypeScript declarations.

JSDoc comments are really only for use by developers, the public documentation is entirely separate.

chrisn:
This is really good. Thank you for making these changes. The only I'm less keen on is https://github.com/bbc/peaks.js/pull/402/commits/5d04a1bbc1ba4f9713689fcd3f16e24283b5c118, because the function is only used in one place.

I'll merge this, then rebase the ES modules (`rollup`) branch again. At that point I want to hold off on further changes until the branch is merged.

jdelStrother:
Thanks!

> The only one I'm less keen on is 5d04a1b, because the function is only used in one place.

I'd really like to make WaveformZoomView.scrollWaveform  part of the public API if possible - ok if I submit a PR with docs & typescript def for it?

chrisn:
Yes, that's OK but I wonder if the parameter should be in seconds rather than pixels, or both even? I'd also want to wait until I merge the ES modules changes.

jdelStrother:
>  I wonder if the parameter should be in seconds rather than pixels, or both even

Could do. Something like this?

```diff
  interface WaveformZoomView extends WaveformView {
    scrollWaveform: (offset: number) => void;
    setStartTime: (time: number) => void;
    setWheelMode: (mode: 'scroll' | 'none') => void;
    setZoom: (options: XOR<{ scale: number | 'auto' }, { seconds: number | 'auto' }>) => void;
+   scrollWaveform: (offset: XOR<{ pixels: number }, { seconds: number }>) => void;
  }
```

chrisn:
Yes. What use case do you have in mind for this? I don't want to add extra complexity, but it would seem that this should be in seconds, as we already have `setStartTime()`.

jdelStrother:
It's for some onscreen buttons that nudge the waveform left & right. For my purposes, I'd prefer pixels rather than time, so that if you're zoomed in, the nudging is more precise time-wise, but still moves the same 'visual' distance.

If you're not sure about it, I can achieve the same thing with setStartTime and some maths, it just seemed like something that would be useful.

chrisn:
My only concern is that someone in future will request adding `scrollWaveform(seconds)` support, so I'd want to make sure the API allows for that, even if it's not implemented initially.

Issue #310, opened 2020-02-04
Modernizing peaks' package distribution


jdelStrother:
Our app depends on waveform-data.js & peaks.js.   Peaks inlines waveform-data into its distributed source file, which would normally result in waveform-data getting duplicated between our app & peaks.  We've been hacking around that for a while by compiling peaks from its src directory:

```js
  resolve: {
    alias: {
      "peaks.js": path.resolve(__dirname, "node_modules/peaks.js/src/main.js")
    }
```

In peaks 0.18 (or maybe 0.18.1), the src directory is no longer included in the npm distribution, so that doesn't seem to be an option any more.

Can we have an npm distribution that includes waveform-data as a peer dependency so we can de-dupe it?

chrisn:
I'm not sure what the right solution is here, to be honest. For now, should I simply add the `src` files back in?

jdelStrother:
I've been testing this some more today... last September, we were seeing waveform-data twice in our source files due to peaks.js inlining it, but that no longer appears to be the case.  I'm not entirely sure if our webpack config changed or peaks did.

So we could just require the peaks.js & waveform-data packages normally, without our webpack hacks.  However, I still think it would be worth including the src files in the package with something like this:

```diff
diff --git a/package.json b/package.json
index 7553e77..05056c7 100644
--- a/package.json
+++ b/package.json
@@ -3,8 +3,10 @@
   "version": "0.18.1",
   "description": "JavaScript UI component for displaying audio waveforms",
   "main": "./peaks.js",
+  "module": "./src/main.js",
   "types": "./peaks.js.d.ts",
   "files": [
+    "src",
     "peaks.js",
     "peaks.js.d.ts"
   ],
```

would would allow webpack & other bundlers to use the `module` entry to compile from source, which seems to do a better job of minifying the resulting package (from 298KB minified javascript to 281KB).

chrisn:
The main thing I changed was to add the `"files"` field, as I noticed the npm package included the demo pages with audio files, making the package much larger than necessary. I also removed the `"browser"` entry in package.json and changed all the AMD dependencies, so I hope this change didn't break your build.

I'm planning to make another change to include only the Konva.js modules that we actually use, rather than the whole bundle. This should also help with file size, so I can also do this at the same time as this fix.

If I create a branch, would you mind testing before I make a new release?

jdelStrother:
Sure, sounds good

chrisn:
[...]

PR #410, opened 2021-07-23
fix bundlesize

adi518:
This PR fixes Rollup's `external` config. It was left an empty array for some strange reason, which means it didn't externalize all production dependencies from the bundle, resulting in a significantly large bundle (`70.4KB`, gzipped, see: https://bundlephobia.com/package/peaks.js@0.26.0). With this fix, the bundlesize is now `16.38KB` (gzipped), a `76.732%` difference. 🤯

Summary:
* Fix Rollup external (impl from TSDX)
* Add `esm` build to enable preliminary tree-shaking.
* Add `analyze` script and `size-limit` (also from TSDX)
* Move build to `dist` folder, which makes it easier to inspect (also from TSDX)
* Fix some lint warnings
* Remove unused dep `watchify`.

~~Btw, as you can see there's an issue with the `waveform-data` dependency, as it unnecessarily includes `core-js`, a set of polyfills that are NOT needed in today's browsers. That alone can be another `15-20KB` reduction for the end-consumer. That should be handled in the dependency repository and later bump its version here.~~

Notes:
* It's probably noteworthy that the library is maintained with npm 7, so anyone with npm <= 6 will regress `package-lock.json` into an old format (`"lockfileVersion": 1`).

image


chrisn:
Thanks, I'll take a look at this. One thing I found is that the demo pages no longer work using the smaller peaks.js bundle.

> Btw, as you can see there's an issue with the `waveform-data` dependency, as it unnecessarily includes `core-js`, a set of polyfills that are NOT needed in today's browsers.

[core.js](https://github.com/bbc/waveform-data.js/blob/752540e52b86f3f3218c50aca38727eb6a24a480/lib/core.js) is part of the waveform-data library, not a separate dependency.

adi518:
I'll check the demo, thanks. Core.js is a bit extraneous, it can originate from the build-tool dependencies or appear as a standalone dependency that's used by the build-tool. It's usually removed by changing the build configuration and then removing it from the dependencies. In this case it's not an explicit dependency, so it only need to be removed by configuration.

adi518:
I fixed the demos. The reason they broke is because of externalising the dependencies obviously, so you have to treat each dome as an app that consumes the library, which means you have to add UMD scripts of the dependencies to each demo HTML page. However, that's tedious and repetitive, so I tricked it. I renamed `build-demo` to what it actually is, `build-custom-markers-demo` and added a new script `build-demo`, which builds the library, but this time, without externalising its dependencies.

chrisn:
> However, that's tedious and repetitive, so I tricked it.

The purpose of the demos is to show people how to use the library. So if we're using external dependencies, the demos should show how to do that.

rollup.config.js
```diff
@@ -24,9 +26,18 @@ export default {
       ]
     }
   ],
-  external: [],
+  // eslint-disable-next-line
+  // https://github.com/formium/tsdx/blob/462af2d002987f985695b98400e0344b8f2754b7/src/createRollupConfig.ts#L63
+  external: (id) => {
+    // bundle in polyfills as TSDX can't (yet) ensure they're installed as deps
+    if (id.startsWith('regenerator-runtime')) {
```
chrisn, line 51: Peaks.js has no dependency on regenerator-runtime, so this isn't needed. The external list can be more simply created using [rollup-plugin-peer-deps-external](https://github.com/pmowrer/rollup-plugin-peer-deps-external).

package.json
```diff
@@ -80,11 +88,12 @@
     "rollup-plugin-terser": "~7.0.2",
     "serve": "~12.0.0",
     "sinon": "~11.1.1",
-    "sinon-chai": "~3.7.0",
-    "watchify": "~4.0.0"
+    "sinon-chai": "~3.7.0"
   },
   "dependencies": {
+    "@size-limit/preset-small-lib": "^5.0.1",
```
chrisn: These should be devDependencies. But in any case, I don't particularly want to add these.

adi518: Outdated change, I already fixed it, but I can remove it.

package.json
```diff
@@ -2,9 +2,15 @@
   "name": "peaks.js",
   "version": "0.26.0",
   "description": "JavaScript UI component for displaying audio waveforms",
-  "main": "peaks.min.js",
-  "module": "./src/main.js",
+  "main": "./dist/peaks.min.js",
+  "module": "./dist/peaks.esm.min.js",
```
chrisn, line 6: I haven't been able to get the esm build to work:

> TypeError: Failed to resolve module specifier "konva/lib/shapes/Line". Relative references must start with either "/", "./", or "../".

.prettierignore
```diff
@@ -0,0 +1 @@
+*
```
chrisn, line 1: This project doesn't use Prettier, so no need to add this file.

package.json
adi518: How did you test it?

chrisn: I use `<script type="module">` in one of the demo pages.

.prettierignore
adi518: I added this to prevent users with Prettier to accidentally reformat the code, which is very different from whatever format you guys are using. Can you tell which formatter to use? I guess you have it installed globally, which means it's not part of the project infra.

chrisn:
Thanks a lot for working on this. Reducing the bundle size is a good idea, and this is really helpful! Using external dependencies it's possible to do this, which overcomes the need to use a module bundler for the custom-markers demo:

```html
<script src="https://unpkg.com/waveform-data@4.1.0/dist/waveform-data.js"></script>
<script src="https://unpkg.com/konva@8.1.0/konva.js"></script>
<script src="/peaks.js"></script>
```

package.json
chrisn: Shouldn't `module` still point to `src/main.js`?

.prettierignore
chrisn: I don't use auto-formatters. I don't really want tooling-specific files in the project, is there no other way to prevent this?

package.json
chrisn: The way I'm thinking about this is: `module` points to `src/main.js` and people who want an ESM should use rollup or webpack or other bundler to resolve all the dependencies.

rollup.config.js
adi518: There's a few plugins that do this, but TSDX's simple function does the same job. We can use either.

adi518:
> Thanks a lot for working on this. Reducing the bundle size is a good idea, and this is really helpful! Using external dependencies it's possible to do this, which overcomes the need to use a module bundler for the custom-markers demo:
> 
> ```
> <script src="https://unpkg.com/waveform-data@4.1.0/dist/waveform-data.js"></script>
> <script src="https://unpkg.com/konva@8.1.0/konva.js"></script>
> <script src="/peaks.js"></script>
> ```
> 
> Using externals for waveform-data and konva is good, but probably not worth it for eventemitter3.

Yes, those are the dependencies that should be added to each demo, if you want to want to showcase usage properly. In a modern app though, it's not needed, so really, every dependency should be externalised.

.prettierignore
adi518: There's a way, but it's slightly verbose. You have to create a new VS Code workspace directory, apply the `.prettierignore` file there, clone the repository and open the workspace in VS Code. It hurts the developer experience. I for one, use "format on save", so I can't apply changes without disabling the formatter first.

chrisn:
But this is why I'm not really seeing much need for this change. Peaks.js is now using ES modules (although the last published release, 0.26.0 doesn't). So, from the next release onwards, anyone using Peaks.js with a modern module bundler will be using the `module` entry in package.json and producing their own bundle, which will de-duplicate any dependencies. 

Using `external` reduces the size of the bundled `peaks.js` file, which only affects people who prefer to use a simple `<script>` tag. For those people, we actually increase the total amount of JavaScript they need to download, because they'll be using the full Konva library, and not just the specific component modules that Peaks.js uses.

package.json
adi518: IIRC, that's the correct way to consume an ESM build with a script tag, I'll have a look at this soon to verify it works.

The `module` entry should point to the ESM build, not the source code, as it is not up to consumers to bundle dependency code. By default, all build tools assume `node_modules` are a set of production-ready dependencies, so all they do is concatenate each dependency bundle into the bundle of your app/library. It's noteworthy that when building a library and externalising its dependencies, all the build-tool does is leave references to whatever each dependency exposes, so when consumed, everything is joined back together without duplications.

adi518:
I explained about ESM and the `module` entry above, but I'll go into further details here. Consumers can't/shouldn't consume your source, because it may include stuff that only works within your development environment, namely you build it in order to provide production-ready code that is decoupled from your development environment. I'll give you an example. Many build-tools provide you with an alias option to resolve dependencies through an alias instead of using relative paths. It usually looks like this:

```js
import Foobar from "@components/Foobar"; // @components = <project-root>/src/components
// instead of
import Foobar from "../../components/Foobar";
```
That won't work if someone were to consumes a dependency from its source code. Although it may not be relevant to this library, you get the idea.

adi518:
@chrisn Any updates?

package.json
chrisn: The ESM build didn't work in the browser because it needs an import map to be able to resolve Konva dependencies, but also other dependencies such as EventEmitter3 and waveform-data aren't ESM format. So the ESM build is only suitable for consumption by build tools.

adi518: I haven't tested the ESM build in the browser yet, but we are **not** replacing UMD with ESM. We build both to cover everything, so I don't see an issue.

chrisn: It was just an observation.

chrisn:
Please take a look at these recent commits on `master`: https://github.com/bbc/peaks.js/commit/0207a452dbba5df8e9edb7c9573e3a0363b181e4 and https://github.com/bbc/peaks.js/commit/1cef03d4e51185424b646b9a20619904ac6bb38d. I have added an ES module build, and added source maps to the ES and UMD modules.

Should we include a minified ES module build? Wouldn't minification be applied by a module build tool that consumes the Peaks.js ES module?

For UMD, I think that a build with all dependencies bundled is still useful, for people who want to simply add a <script> tag to their page. The custom marker feature [won't work](https://github.com/bbc/peaks.js/issues/310#issuecomment-855465401), so it would be good if this build could disable that feature, but I don't think that's essential (I can document this limitation). In addition, we'd provide a UMD build with Konva.js and waveform-data as external dependencies, for people who want to use a <script> tag but also want to use custom markers.

chrisn:
I have just released v0.27.0 that includes two UMD builds (one that has all dependencies bundled, the other without Konva and waveform-data), and an ES module build. Feedback is welcome on how well this works. I'll close this issue, so please raise new issues if you encounter any difficulties. Thanks

adi518:
Cool, I'll check it out and provide feedback, if necessary. 👍

PR #417, opened 2021-11-05
Add logger to options

rowild:
It seems that without `logger` defined in the `options` object, an error is thrown:

```sh
Uncaught TypeError: this._peaks.logger is not a function
    at WaveformZoomView.setZoom (peaks.js:14546)
    at zoomable-waveform.html:160
    at peaks.js:16876
    at peaks.js:16162
    at Worker.listener (peaks.js:16125)
```

Adding a `logger` definition makes the example work again.

Before submitting a pull request, please read our [contributor guidelines](https://github.com/bbc/peaks.js/blob/master/CONTRIBUTING.md).

Pull requests that don't follow the guidelines risk not being accepted.

### DIFF
--- a/demo/zoomable-waveform.html
+++ b/demo/zoomable-waveform.html
@@ -139,7 +139,8 @@
             multiChannel: false
           },
           zoomLevels: [128],
-          keyboard: true
+          keyboard: true,
+          logger: console.error.bind(console)
         };
 
         Peaks.init(options, function(err, peaksInstance) {
--- a/demo/zoomable-waveform.html
+++ b/demo/zoomable-waveform.html
@@ -139,8 +139,7 @@
             multiChannel: false
           },
           zoomLevels: [128],
-          keyboard: true,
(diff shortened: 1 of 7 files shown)


rowild:
Converted to draft, because I wonder if this is a use case, for which a test would be helpful...

chrisn:
Sorry if this is a newly introduced bug. I haven't looked into it yet, but a test case to reproduce it would be welcome.

rowild:
@chrisn No problem! (Not sure if it is new, I think it didn't work before, either...) However, the problem is me and tests. I simply do not know how to write one. I must leave this decision to you – I am very sorry about not being able to help here!

chrisn:
I have fixed the bug and added some test cases. The issue was that the logger was not being correctly initialised. It shouldn't be necessary to provide a `logger` config option, so I removed that change.

rowild:
Thank you very much, @chrisn !

PR #418, opened 2021-11-06
Fix a TOC jump link

rowild:
Rename "ES module bundle" to "Install to npm" to make the TOC jump link work. 

I decided to rename the header of the documentation chapter, since a further explaining text uses "ES module bundler" anyway.


chrisn:
Perfect. Thank you!

PR #435, opened 2022-01-14
feat: add contextmenu to point-marker and point-layer

rowild:
Before submitting a pull request, please read our [contributor guidelines](https://github.com/bbc/peaks.js/blob/master/CONTRIBUTING.md).

Pull requests that don't follow the guidelines risk not being accepted.


rowild:
Hi Chris!

I haven't contacted you about that directly and am (at least) therefore failing the contributing guide. But I thought maybe the  Draft feature would do a similar thing and inform you about this idea... 

My thought:
It would be great to also have "contextmenu" available as a point-marker event. That would enable it to build a custom right-click dropdown menu on a pointer.

If this is something that you would accept, it would probably need implementation in segments and the waveforms, as well.

I am a bit shocked that that all checks failed. I was careful about code formatting and code structure... what am I doing wrong?

BTW: all in good time!

chrisn:
Exposing contextmenu events sounds like a good addition.

I think the PR check fails because your commit deletes package-lock.json.

PR #448, opened 2022-03-30
Fix documentation showAxisLabels

tdurand:
Hi ! Awesome lib thanks !

Seems there is a mistake in the documentation, the option is `showAxisLabels` and not `showAxisTimeLabels`

Keep up the great work! 

Thibault


chrisn:
Thank you!

PR #453, opened 2022-05-01
label waveformCache prop as optional

tscz:
This PR fixes a little typo in the typescript definition: The waveform cache is currently not labelled as optional in typescript, even though it is intended to be optional.


chrisn:
Sorry, I missed that. Thank you!

PR #494, opened 2023-06-05
Relax konva version range

jdelStrother:
Konva 9.0 / 9.1 have been working fine for us with peaks. How about relaxing the version range?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,7 @@
         "node": ">= 8.11.2"
       },
       "peerDependencies": {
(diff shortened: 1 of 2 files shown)


jdelStrother:
Yeah, the commonjs switch seemed a bit of a regression - the only real discussion I found on it was here: https://discord.com/channels/753274212814422176/755201767934918827/threads/1096466627287060592. But in practice it only increased our bundle size by 1.5kb, not noticed any other issues in our webpack-based build with 9.x.

Issue #507, opened 2023-10-03
Playhead text background color


amber-huo:
Hi, I would like to be able to customize a background/highlighted color to the playhead text. If the waveform and playhead text happen to be similarly colored, the text doesn't show up very well when it overlaps high amplitude areas. So it's for better color contrast/generally more visible text. Something like:
![image](https://github.com/bbc/peaks.js/assets/105802812/efa5bac2-9c63-4420-92c7-616d2f327e29)

chrisn:
Thanks for this suggestion. It sounds good, we could add an option for the background color and possibly also for padding around the text.

amber-huo:
Thanks @chrisn, I opened a PR for both background color and padding around the text, like so:
![image](https://github.com/bbc/peaks.js/assets/105802812/4265b3fa-d409-4376-b187-69430064862b)

chrisn:
Your PR is now merged. Thanks again for contributing! 🙏

chrisn:
@amber-huo I have just released v3.2.0 which includes your PR.

PR #508, opened 2023-10-04
Playhead text customization

amber-huo:
#507 

This adds two options for customizing playhead text:
- background color
- padding around the text


chrisn:
Thank you! With the padding option, could this apply to both the left and right of the text, at the moment it looks to be only on the right.  What do you think?

Another thing I would suggest is rename the `playheadBkgrdColor` as `playheadBackgroudColor`, so it's not abbreviated.

amber-huo:
> Thank you! With the padding option, could this apply to both the left and right of the text, at the moment it looks to be only on the right. What do you think?
> 
> Another thing I would suggest is rename the `playheadBkgrdColor` as `playheadBackgroudColor`, so it's not abbreviated.

Sure, took both your suggestions.

chrisn:
Thank you! I'm just doing some testing. There are a couple of other minor things I'd like to consider. One is, when the playhead is close to the right-hand edge of the view, the playhead text moves to the left of the playhead, so ideally this would also take into account the padding. The other is that there's a really small gap between the playhead and the padding, would it look better if we could remove this?

amber-huo:
Ok, I adjusted for padding when the text moves to the left, and removed the gap. The padding has a default of 2 so the text should look the same as before when there is no custom background or padding set.

chrisn:
This looks great, thank you!

chrisn:
The only final things are to add some documentation and a few really minor nitpicks on the code. Would you prefer if I do those, or should I suggest changes?

amber-huo:
I'm cool if you go ahead and do those! Thanks.

PR #518, opened 2023-12-04
playheadPadding should be optional in ViewOptions

jdelStrother:
This is defaulted to 2 in defaultViewOptions, so as far as I can tell should be an optional property here?


chrisn:
Thank you, yes, this should have been optional.

Issue #350, opened 2020-12-19
FR: Fixed position on zoomed waveform


Christilut:
My goal is to create a waveform display similar to what DJ apps have and Peaks.js is almost ideal for this. The overview synced with the zoomed waveform is great.
It would be really great if I can somehow get the zoomed waveform to be in a fixed position at the center and let the waveform itself move. The zoomed waveform indicator would always point to the current position in time.

Picture for reference:
![image](https://user-images.githubusercontent.com/3471363/102695100-8eaf2c00-4225-11eb-8159-60ae20a3c00d.png)

The zoomed waveform indicator would not move at all. The waveform would move below it. The overview waveform behaves as normal.

I'm ready to do my best for a PR but I'd love some pointers!

chrisn:
I can see that some refactoring would be needed to add this feature... The main things that would need to change are:

* [`WaveformZoomView._syncPlayhead`](https://github.com/bbc/peaks.js/blob/c2e5517de4677fa04865c2e9f98b94735655730f/src/waveform-zoomview.js#L269), which is responsible for scrolling the waveform when the playhead reaches the right-hand side of the view
* The `WaveformZoomView`'s [`MouseDragHandler`](https://github.com/bbc/peaks.js/blob/c2e5517de4677fa04865c2e9f98b94735655730f/src/waveform-zoomview.js#L129), which responds to click and drag events to scroll the waveform
* The playhead animation in [`PlayheadLayer`](https://github.com/bbc/peaks.js/blob/c2e5517de4677fa04865c2e9f98b94735655730f/src/playhead-layer.js#L245), which updates the playhead position during playback
* [`PlayheadLayer.updatePlayheadTime`](https://github.com/bbc/peaks.js/blob/c2e5517de4677fa04865c2e9f98b94735655730f/src/playhead-layer.js#L163), which updates the playhead position, e.g., after the user changes the zoom level

jtag05:
Is there still the possibility of implementing this? I'm coming over from wavesurfer and having the fixed playhead feels more natural.

chrisn:
Welcome! Yes, I'm interested to implement this, but can't give any promises when.

oenu:
Commenting to indicate interest in this feature with caveat that I do not have enough free time to contribute code rn

PR #558, opened 2025-01-15
Add initial fixed playhead implementation

TristanCacqueray:
This is a first attempt at implementing #350


TristanCacqueray:
Hello @chrisn , I hope it's ok I took a stab at implementing this feature. There is still a bug when deactivating the feature during playback, in that case the waveform is not updated properly. I guess that's because this still needs the other changes you mentioned in the issue. However, if you leave the feature enabled, then it works as expected.

Though the canvas movement is not very smooth, there a small but noticeable stutter as the waveform is moved instead of the playhead. Is there something that should be done about it?

chrisn:
Thank you! Sorry for taking so long to reply, I'll try to find time to look into this soon.

PR #564, opened 2025-07-28
Updated TypeScript declarations

tnibler:
This change just adds the `zoomview.update` event to the typescript declarations.

### DIFF
--- a/peaks.js.d.ts
+++ b/peaks.js.d.ts
@@ -415,6 +415,11 @@ declare module 'peaks.js' {
     previousZoom: number;
   }
 
+  interface ZoomViewUpdateEvent {
+    startTime: number;
+    endTime: number;
+  }
+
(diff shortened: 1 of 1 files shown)


peaks.js.d.ts
```diff
@@ -415,6 +415,11 @@ declare module 'peaks.js' {
     previousZoom: number;
   }
 
+  interface ZoomViewChangeEvent {
```
chrisn, line 418: I think I'd prefer to name this `ZoomViewUpdateEvent`.

chrisn:
Thanks! I don't use TypeScript so updating the declarations sometimes gets overlooked, so this is very welcome.

chrisn (approved):

tnibler:
Thank you for the quick responses :) Love the library!

PR #571, opened 2025-11-07
Relax konva version range

jdelStrother:
Konva 10 seems to work fine with peaks, how about relaxing the version range in peerDependencies?


chrisn:
Thanks for testing this. I've merged the PR in this project's [new home](https://codeberg.org/chrisn/peaks.js).

